PR #2, opened 2018-03-21
VLCRendererDiscoverer: Track discovered renderer items

Mikanbu:
Creating a new instance of `VLCRendererItem` each time we got an event
from libvlc led to duplication and issues with item deletions. Thus,
this patch checks before creating a new instance, if a `VLCRendererItem`
exist with a `libvlc_renderer_item_t *`.


Sources/VLCRendererDiscoverer.m
```diff
@@ -153,16 +154,37 @@ - (void)dealloc
...
 }
 
+- (VLCRendererItem *)discoveredItemsContainsItem:(libvlc_renderer_item_t *)item
+{
+    for (VLCRendererItem *tmpItem in _items) {
+        if (tmpItem.rendererItem == item) {
+            return tmpItem;
+        }
+    }
+    return nil;
```
carolanitz: what you could do is to have a method called addRendererItemIfNeeded

-(void) addRendererItemIfNeeded:(libvlc_renderer_item_t *)item {
  for (VLCRendererItem *tmpItem in _items) {
         if (tmpItem.rendererItem == item) {
             return;
         }
     }
   VLCRenderItem *tmpItem =  [[VLCRendererItem alloc] initWithCItem: pointerValue];
   [_items addObject: tmpItem];
   [_delegate rendererDiscovererItemAdded:self item:tmp];
}

Sources/VLCRendererDiscoverer.m
```diff
@@ -153,16 +154,37 @@ - (void)dealloc
...
 
-- (void)itemDeleted:(VLCRendererItem *)item
+- (void)itemDeleted:(NSValue *)item
 {
-    [_delegate rendererDiscovererItemDeleted:self item:item];
+    libvlc_renderer_item_t *pointerValue = item.pointerValue;
+    VLCRendererItem *tmp = [self discoveredItemsContainsItem:pointerValue];
+
+    [_items removeObject:tmp];
+    [_delegate rendererDiscovererItemDeleted:self item:tmp];
```
carolanitz: This would turn into : 
 for (VLCRendererItem *tmpItem in _items) {
         if (tmpItem.rendererItem == item.pointerValue) {
             [_items removeObject: tmpItem];
             [_delegate rendererDiscovererItemDeleted:self item: tmpItem];
             return; 
         }
     }

Though double check that you don't get an issue with modified while being iterated

Sources/VLCRendererDiscoverer.m
```diff
@@ -153,16 +154,37 @@ - (void)dealloc
...
+- (void)itemAdded:(NSValue *)item
 {
-    [_delegate rendererDiscovererItemAdded:self item:item];
+    libvlc_renderer_item_t *pointerValue = item.pointerValue;
+    VLCRendererItem *tmp = [self discoveredItemsContainsItem:pointerValue];
+
+    if (!tmp) {
+        tmp = [[VLCRendererItem alloc] initWithCItem: pointerValue];
+        [_items addObject:tmp];
+    }
```
carolanitz: then you can reduce this here to 
[self addRendererItemIfNeeded:item];
and only ever call the delegate if an item was really added

carolanitz (changes requested):
I suggested a way how this code could be a bit clearer and you can avoid the delegate calls with nil :)

Mikanbu:
Thank you for the review! I updated my patches but I kept the method to check if the item existed to avoid a loop duplication in the add & delete

Sources/VLCRendererDiscoverer.m
```diff
@@ -153,16 +154,45 @@ - (void)dealloc
...
+
 #pragma mark - Handling libvlc event callbacks
 
-- (void)itemAdded:(VLCRendererItem *)item
+- (void)itemAdded:(NSValue *)item
 {
-    [_delegate rendererDiscovererItemAdded:self item:item];
+    libvlc_renderer_item_t *pointerValue = item.pointerValue;
+    VLCRendererItem *tmp = [self discoveredItemsContainsItem:pointerValue];
+
```
fkuehne: Consider renaming both variables to something more self-explaining so the code becomes more readable.

Sources/VLCRendererDiscoverer.m
```diff
@@ -153,16 +154,45 @@ - (void)dealloc
...
+    }
 }
 
-- (void)itemDeleted:(VLCRendererItem *)item
+- (void)itemDeleted:(NSValue *)item
 {
-    [_delegate rendererDiscovererItemDeleted:self item:item];
+    libvlc_renderer_item_t *pointerValue = item.pointerValue;
+    VLCRendererItem *tmp = [self discoveredItemsContainsItem:pointerValue];
+
```
fkuehne: Same here.

Sources/VLCRendererDiscoverer.m
```diff
@@ -153,16 +154,45 @@ - (void)dealloc
     return [list copy];
 }
 
+- (VLCRendererItem *)discoveredItemsContainsItem:(libvlc_renderer_item_t *)item
```
fkuehne: Minor grammar nag nag:
discoveredItemsContainsItem -> discoveredItemsContainItem

fkuehne (changes requested):
Great work! Minor nitpicks ;)

Mikanbu:
Thanks for the review, PR updated!

Headers/Public/VLCRendererDiscoverer.h
```diff
@@ -72,6 +72,11 @@
  */
 @property (nonatomic, readonly, copy) NSString * _Nonnull name;
 
+/**
+ * Renderers of the discoverer
+ */
+@property (nonatomic, readonly, copy) NSArray<VLCRendererItem *> * _Nonnull renderers;
```
carolanitz: actually you can just add NS_ASSUME_NONNULL_BEGIN to the header and delete the nonnull everywhere and that will help with readability :) 
https://developer.apple.com/swift/blog/?id=25

Sources/VLCRendererDiscoverer.m
```diff
@@ -29,6 +29,7 @@
 @interface VLCRendererDiscoverer()
 {
     libvlc_renderer_discoverer_t *_rendererDiscoverer;
+    NSMutableArray<VLCRendererItem *> *_items;
```
carolanitz: rendererItems / discoveredItems ?

Sources/VLCRendererDiscoverer.m
```diff
@@ -153,16 +154,45 @@ - (void)dealloc
     return [list copy];
 }
 
+- (VLCRendererItem *)discoveredItemsContainItem:(libvlc_renderer_item_t *)item
+{
+    for (VLCRendererItem *tmpItem in _items) {
+        if (!strcmp(libvlc_renderer_item_name(tmpItem.rendererItem), libvlc_renderer_item_name(item))
+            && !strcmp(libvlc_renderer_item_type(tmpItem.rendererItem), libvlc_renderer_item_type(item))) {
```
carolanitz: this was very hard to read and I had a hard time seeing what is actually being checked for. 

Could you maybe write this instead?
Bool hasSameName = !strcmp(libvlc_renderer_item_name(tmpItem.rendererItem), libvlc_renderer_item_name(item));
Bool hasSameType = !strcmp(libvlc_renderer_item_type(tmpItem.rendererItem), libvlc_renderer_item_type(item));
if (hasSameName && hasSameType)

Sources/VLCRendererDiscoverer.m
```diff
@@ -153,16 +154,45 @@ - (void)dealloc
     return [list copy];
 }
 
+- (VLCRendererItem *)discoveredItemsContainItem:(libvlc_renderer_item_t *)item
+{
+    for (VLCRendererItem *tmpItem in _items) {
```
carolanitz: *tmpItem in _items -> *discoveredItem in _discoveredItems

Sources/VLCRendererDiscoverer.m
```diff
@@ -153,16 +154,45 @@ - (void)dealloc
...
-- (void)itemDeleted:(VLCRendererItem *)item
+- (void)itemDeleted:(NSValue *)item
 {
-    [_delegate rendererDiscovererItemDeleted:self item:item];
+    libvlc_renderer_item_t *renderer_item = item.pointerValue;
+    VLCRendererItem *rendererItem = [self discoveredItemsContainItem:renderer_item];
+
+    if (rendererItem) {
+        [_items removeObject:rendererItem];
+        [_delegate rendererDiscovererItemDeleted:self item:rendererItem];
```
carolanitz: 👍

Sources/VLCRendererDiscoverer.m
```diff
@@ -153,16 +154,45 @@ - (void)dealloc
...
+- (void)itemAdded:(NSValue *)item
 {
-    [_delegate rendererDiscovererItemAdded:self item:item];
+    libvlc_renderer_item_t *renderer_item = item.pointerValue;
+    VLCRendererItem *rendererItem = [self discoveredItemsContainItem:renderer_item];
+
+    if (!rendererItem) {
+        rendererItem = [[VLCRendererItem alloc] initWithCItem:renderer_item];
+        [_items addObject:rendererItem];
+        [_delegate rendererDiscovererItemAdded:self item:rendererItem];
```
carolanitz: :+1:

Sources/VLCRendererDiscoverer.m
```diff
@@ -39,8 +40,8 @@ static void HandleRendererDiscovererItemAdded(const libvlc_event_t *event, void
     @autoreleasepool {
         [[VLCEventManager sharedManager] callOnMainThreadObject:(__bridge id)(self)
                                                      withMethod:@selector(itemAdded:)
-                                           withArgumentAsObject:[[VLCRendererItem alloc] initWithCItem:
-                                                                 event->u.renderer_discoverer_item_added.item]];
+                                           withArgumentAsObject:[NSValue valueWithPointer:
+                                                                 event->u.renderer_discoverer_item_added.item]] ;
```
carolanitz: what is this u btw? That's not very discriptive

carolanitz (changes requested):
through the new strcmp it got a bit unreadable and I noticed two things that I missed in the first review. Otherwise it looks good :)

Headers/Public/VLCRendererDiscoverer.h
Mikanbu: If I add `NS_ASSUME_NONNULL_BEGIN` wouldn't that mean that there will be multiple blocks since there are some element that are nullable such as the delegate?
I personally prefer stating explicitly that an element is `Nonnull` or `Nullable`.

Sources/VLCRendererDiscoverer.m
Mikanbu: If I'm not mistaken, it's a union defined in libvlc.

Sources/VLCRendererDiscoverer.m
Mikanbu: I agree that it isn't the most readable code here, but it is C and I don't know if creating two variables will be worth in this case?

Headers/Public/VLCRendererDiscoverer.h
carolanitz: there won't be mutiple blocks. It just assumes everything is nonnull unless stated differently. You can take a look at the NSString class to see how Apple does it

Headers/Internal/VLCRendererItem+Init.h
```diff
@@ -41,6 +41,6 @@
  * Returns the C renderer item
  * \return Renderer item
  */
-- (libvlc_renderer_item_t * _Nonnull)rendererItem;
```
carolanitz: I understand why you did that but since this is part of your wrapper and VLCKit the CamelCase is and was absolutely correct here

Mikanbu: Even though it isn't exposed to the users?
My reasoning behind it was:
- rendererItem name conflict in the strcmp loop(VLCRendererDiscoverer, line 160)
- It's a c pointer

carolanitz: Good Points. The naming conflict should be fine since it's a property on the rendereritem but I see how that could be confusing. For the c pointer argument, I'm trying to see if we have that case somewhere else in VLCkit and how we handled that or try to find Apple code

carolanitz: I think the libVLCMediaDescriptor on VLCMedia might be a comparable example. We have in the past used void * for any c pointers. You can find the same behavior in VLCMediaPlayer with the internal variable _playerInstance.
prefixing it with libVLC would also resolve the naming conflict here

Mikanbu: I don't mind renaming it but isn't `rendererItem.libvlcRendererItem` a little verbosy? 😄

carolanitz: Lololol :D may I remind you that you're coding in Objective-C ? :D That's nothing!
*Looks over to CMAttitudeReferenceFrameXArbitraryCorrectedZVertical*

Mikanbu: lol that's a very good point!

fkuehne: Sorry for being late to the party, but using a void pointer would be the correct way to do it.

Mikanbu: Even if we're not exposing it?

fkuehne (approved):

fkuehne:
Please consider squashing some of the commits before merging it upstream, so we don't have commits removing single spaces in the commit history.

Headers/Internal/VLCRendererItem+Init.h
fkuehne: That would follow the logic of the internal VLCLibVLCbridging.h header, yes.

Mikanbu:
Updated, now uses `VLCLibVLCbridging` logic.

Headers/Internal/VLCLibVLCBridging.h
```diff
@@ -146,6 +147,27 @@
...
+ */
+@interface VLCRendererItem (VLCRendererItemBridging)
+/**
+ * Initializer method to create an VLCRendererItem with an `libvlc_renderer_item_t *`.
+ *
+ * \param renderer item.
+ * \note This initializer is not meant to be used externally.
+ * \return An instance of `VLCRendererItem`, can be nil.
+ */
+- (instancetype)initWithRendererItem:(libvlc_renderer_item_t *)item;
```
fkuehne: Can we please have a void pointer here to match the rest of the file?

Sources/VLCRendererItem.m
```diff
@@ -40,9 +40,9 @@ - (void)dealloc
 
 @end
 
-@implementation VLCRendererItem (Internal)
+@implementation VLCRendererItem (VLCRendererItemBridging)
 
-- (instancetype)initWithCItem:(libvlc_renderer_item_t *)item
+- (instancetype)initWithRendererItem:(libvlc_renderer_item_t *)item
 {
```
fkuehne: Same here.

fkuehne (changes requested):
Minor nit-pick, then we are good to go.

Please squash most of the commits so we don't 20+ commits for this to the history.

fkuehne (approved):

carolanitz (approved):

Mikanbu:
Merged, thanks @fkuehne and @carolanitz for the reviews ❤️

PR #6, opened 2018-04-24
VLCMediaPlayer: Don't release equalizer on disable

Mikanbu:
Remove the release of the equalizer instance when disabling it.
Releasing it every time could lead to some information loss.

`_equalizerInstance` is released on `dealloc`.

### DIFF
--- a/Sources/VLCMediaPlayer.m
+++ b/Sources/VLCMediaPlayer.m
@@ -974,17 +974,11 @@ static void HandleMediaPlayerSnapshot(const libvlc_event_t * event, void * self)
 - (void)setEqualizerEnabled:(BOOL)equalizerEnabled
 {
     _equalizerEnabled = equalizerEnabled;
-    if (!_equalizerEnabled) {
-        libvlc_media_player_set_equalizer(_playerInstance, NULL);
-
-        if (_equalizerInstance)
-            libvlc_audio_equalizer_release(_equalizerInstance);
-        return;
-    }
-
-    if (!_equalizerInstance)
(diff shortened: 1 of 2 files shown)


Sources/VLCMediaPlayer.m
```diff
@@ -974,17 +974,11 @@ - (NSInteger)currentAudioPlaybackDelay
...
-        libvlc_media_player_set_equalizer(_playerInstance, NULL);
-
-        if (_equalizerInstance)
-            libvlc_audio_equalizer_release(_equalizerInstance);
-        return;
-    }
-
-    if (!_equalizerInstance)
+    if (!_equalizerInstance) {
         _equalizerInstance = libvlc_audio_equalizer_new();
```
carolanitz: this can return NULL and should be handled

carolanitz (changes requested):

Sources/VLCMediaPlayer.m
Mikanbu: You're right, I totally missed it!
I'm not sure how it should be handled? Since it's a setter it feels wrong to have a return value.
Would an NSAssert suffice? 🤔

carolanitz: Ideally I would change the return value but that would break api. I think not setting the equalizer to enabled and handle that case in iOS and change the description to mention that would be good.

fkuehne (approved):
LGTM.

carolanitz (approved):

Mikanbu:
Merged, thanks for the reviews!

PR #7, opened 2018-04-24
Expose viewpoint values

carolanitz:


Sources/VLCMediaPlayer.m
```diff
@@ -1104,20 +1104,44 @@ - (void)stop
...
-    _viewpoint->f_yaw = yaw;
-    _viewpoint->f_pitch = pitch;
-    _viewpoint->f_roll = roll;
-    _viewpoint->f_field_of_view = fov;
+- (BOOL)updateViewpoint:(CGFloat)yaw pitch:(CGFloat)pitch roll:(CGFloat)roll fov:(CGFloat)fov absolute:(BOOL)absolute
+{
+    [self viewPoint]->f_yaw = yaw;
+    [self viewPoint]->f_pitch = pitch;
+    [self viewPoint]->f_roll = roll;
+    [self viewPoint]->f_field_of_view = fov;
```
fkuehne: Those four setters are not safe for 32bit as CGFloat is 64bit for 64bit architectures while 32bit for 32bit architectures. VLC's variables however are always 32bit.

Headers/Public/VLCMediaPlayer.h
```diff
@@ -754,17 +754,45 @@ extern NSString *const VLCTitleDescriptionIsMenu;
...
+
+/**
+ * Get the view point roll in degrees
+ *
+ * \return view point roll in degrees ]-180;180]
+ */
+@property (nonatomic, assign, readonly) CGFloat roll;
+
+/**
+ * Set/Get the adjust filter's gamma value
```
fkuehne: ehm.

Headers/Public/VLCMediaPlayer.h
```diff
@@ -754,17 +754,45 @@ extern NSString *const VLCTitleDescriptionIsMenu;
...
 /**
  * Updates viewpoint with given values.
- * \param yaw new yaw.
- * \param pitch new pitch.
- * \param roll new roll.
- * \param fov new field of view.
+ * \param yaw new yaw.            view point yaw in degrees  ]-180;180]
+ * \param pitch new pitch.        view point pitch in degrees  ]-90;90]
+ * \param roll new roll.          view point roll in degrees ]-180;180]
+ * \param fov new field of view.  field of view in degrees ]0;180[ (default 80.)
```
fkuehne: I would totally remove the old comments and just use your new ones as those are lot more understandable.

fkuehne (changes requested):
Minor nit-picks.

Sources/VLCMediaPlayer.m
chouquette: Hm, are you sure? VLC's integer variables are int64_t

fkuehne: Sure, for integer there is no such problem. However, https://www.videolan.org/developers/vlc/doc/doxygen/html/structlibvlc__video__viewpoint__t.html states simply "float".

Sources/VLCMediaPlayer.m
```diff
@@ -1108,7 +1108,6 @@ - (libvlc_video_viewpoint_t *)viewPoint
 {
     if (_viewpoint == NULL) {
         _viewpoint = libvlc_video_new_viewpoint();
-        NSAssert(_viewpoint != nil, @"Viewpoint wasn't allocated");
```
tguillem: indeed, alloc errors should not be checked with an assert

Sources/VLCMediaPlayer.m
```diff
@@ -1108,40 +1108,54 @@ - (libvlc_video_viewpoint_t *)viewPoint
 {
     if (_viewpoint == NULL) {
         _viewpoint = libvlc_video_new_viewpoint();
```
tguillem, line 1110: You should maybe create this variable from the VLCMediaPlayer constructor, and handle the alloc fail here.
That way, you are sure that _viewpoint can't be NULL during the live cycle of the object.

carolanitz: but it's still fine if it is. You just can't pan or change the viewpoint

tguillem: Yes it is fine. But when possible, I prefer to initialize/allocate small structures/classes in the init function. That way, you don't have to handle the case where the object is null in the other part of code. This make the code simpler IMHO.

carolanitz:
this was merged with 3650673ac6cd121729c15b131c6d070e31f67bbc

PR #8, opened 2018-05-03
build: Change modules for Chromecast

Mikanbu:
This patch removes some modules from the blacklist and enables modules needed for Chromecast.

### DIFF
--- a/compileAndBuildVLCKit.sh
+++ b/compileAndBuildVLCKit.sh
@@ -323,7 +323,7 @@ buildLibVLC() {
     export USE_FFMPEG=1
     ../bootstrap ${BUILD} --host=${TARGET} --prefix=${VLCROOT}/contrib/${OSSTYLE}-${TARGET}-${ARCH} --disable-gpl \
         --enable-ad-clauses \
-        --disable-disc --disable-sout \
+        --disable-disc \
         --disable-sdl \
         --disable-SDL_image \
         --disable-iconv \
@@ -356,7 +356,6 @@ buildLibVLC() {
         --disable-gnutls \
         --disable-lua \
         --disable-luac \
-        --disable-protobuf \
         --disable-aribb24 \
         --disable-aribb25 \
         --enable-vpx \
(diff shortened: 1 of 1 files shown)


compileAndBuildVLCKit.sh
```diff
@@ -547,10 +546,7 @@ buildLibVLC() {
     demuxdump
     fingerprinter
     output_udp
-    output_http
     output_livehttp
-    libmux
-    stream_out
```
fkuehne: Please add more precise removal rules instead. We definitely don't need all stream_out and all muxing modules.

compileAndBuildVLCKit.sh
```diff
@@ -356,7 +356,6 @@ buildLibVLC() {
         --disable-gnutls \
         --disable-lua \
         --disable-luac \
-        --disable-protobuf \
```
fkuehne, line 359: Does this compile for all platforms, notably including tvOS?

fkuehne (changes requested):

compileAndBuildVLCKit.sh
Mikanbu: Tested with tvOS, nobody is complaining!

Mikanbu:
@fkuehne Updated~

compileAndBuildVLCKit.sh
```diff
@@ -547,6 +547,22 @@ buildLibVLC() {
...
+    stream_out_dummy
+    stream_out_duplicate
+    stream_out_es
+    stream_out_gather
+    stream_out_mosaic_bridge
+    stream_out_record
+    stream_out_rtp
+    stream_out_setid
+    stream_out_smem
+    stream_out_stats
```
tguillem: I would prefer enabling all stream_out plugins.
I'm afraid we'll loose time when we will be wondering why some sout module are not working. Furthermore, these modules are very light.

Mikanbu:
Merged.

PR #10, opened 2018-05-14
Add singular form for verbose time description

mkchoi212:
When using `VLCTime.verboseDescription`,

1 **minutes** 20 seconds is now 1 **minute** 20 seconds

### DIFF
--- a/Sources/VLCTime.m
+++ b/Sources/VLCTime.m
@@ -114,21 +114,15 @@
     long mins = (long)((positiveDuration / 60) % 60);
     long seconds = (long)(positiveDuration % 60);
     BOOL remaining = duration < 0;
-    NSString *format;
-    if (hours > 0) {
-        format = remaining ? NSLocalizedString(@"%ld hours %ld minutes remaining", nil) : NSLocalizedString(@"%ld hours %ld minutes", nil);
-        return [NSString stringWithFormat:format, hours, mins, remaining];
-    }
-    if (mins > 5) {
-        format = remaining ? NSLocalizedString(@"%ld minutes remaining", nil) : NSLocalizedString(@"%ld minutes", nil);
-        return [NSString stringWithFormat:format, mins, remaining];
-    }
-    if (mins > 0) {
-        format = remaining ? NSLocalizedString(@"%ld minutes %ld seconds remaining", nil) : NSLocalizedString(@"%ld minutes %ld seconds", nil);
-        return [NSString stringWithFormat:format, mins, seconds, remaining];
-    }
-    format = remaining ? NSLocalizedString(@"%ld seconds remaining", nil) : NSLocalizedString(@"%ld seconds", nil);
-    return [NSString stringWithFormat:format, seconds, remaining];
+   
+    NSDateComponents *components = [[NSDateComponents alloc] init];
+    [components setHour:hours];
+    [components setMinute:mins];
+    [components setSecond:seconds];
+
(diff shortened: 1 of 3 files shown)


fkuehne (approved):
Patch itself looks good to me!

What I’m wondering about - didn’t newer versions of macOS / iOS introduce convenience code to print such user visible strings in all locales? Can we use that?

mkchoi212:
@fkuehne Are you referring to [this](http://macoscope.com/blog/effective-localization-when-working-with-language-plural-rules/)?

It's cool but I don't see any `Localizable.strings` files in the project. Am I missing something?

carolanitz:
awesome!!! This is exactly the reason why we need the Testsuite. To find all those tiny bugs and fix them along the way and make sure that this doesn't break in the future 🙌

carolanitz (approved):

carolanitz:
@fkuehne I think you might be talking about using the Date components Api. We should definitely switch to that here and not try to reimplement the logic. 
I'm not sure if there are differences in time formats by country but if, we don't handle that. 
Also we would be wrong for 1 hour and 1 second.
The code should get more consize by using date components but make sure to only instantiate the object once for the entire class since it's very expensive to create.

https://developer.apple.com/documentation/foundation/datecomponents

fkuehne:
Yes, that’s exactly the API I had in mind. Since we will require macOS 10.10 for VLCKit 4, it will be a good idea to switch to it.

mkchoi212:
One question! When you have 1:00:30, do we want this to be "1 hour 30 seconds" or just "1 hour"?

fkuehne:
I would prefer “1 hour 30 seconds” to be as precise as possible. Custom strings can be created easily by the client app if desired.

mkchoi212:
I also noticed "1 hour 20 minutes 30 seconds" is represented as "1 hour 20 minutes". Is this desired behavior?

carolanitz:
Why don't you take a look at what the nsdateformatter will give you by default? The rest just comes down to personal preference  ¯\_(ツ)_/¯  and like felix says if someone wants something else they can easily adapt it

mkchoi212:
Ok cool. I'll go with the default format. I was just wondering cause I thought there was some particular reason behind things :D

Sources/VLCTime.m
```diff
@@ -24,7 +25,18 @@
...
+- (void)initDateComponents;
+@end
+
 @implementation VLCTime
+static NSDateComponents *components = nil;
+
+- (void)initDateComponents
+{
+    if (!components)
+        components = [[NSDateComponents alloc] init];
```
carolanitz: turns out I've been mixing up the expensiveness of NSDateFormatter with NSDateComponents.
You can use this one actually directly.

carolanitz (approved):
the rest looks good. I would add a couple of examples in the description and update the News to note that we adjusted this method . As reference for the description I would point to - (NSString *)stringFromDate:(NSDate *)fromDate toDate:(NSDate *)toDate; and I think this can be merged then

Sources/VLCTime.m
mkchoi212: 👍

carolanitz:
I just tested it and it works great but the localization for remaining doesn't work. I will merge it anyway because it's improvement.
For german you get 1 Stunde 33 Minuten und 41 Sekunden remaining that should be fixed at some point

carolanitz:
this was merged

PR #14, opened 2018-05-23
Create Rake File Tasks

mkchoi212:
`fastlane` folder has been created with a simple lane that runs tests for MobileVLCKit; which is empty as of right now :p


carolanitz:
I think for MobileVLCKit we don't need to use fastlane and we can work directly with xcodebuild in the circleci yaml

mkchoi212:
So instead of using fastlane, I've decided to use rake in order to organize all the required commands to build/test VLCKit; I was looking around the web and got inspired by the folks at [PSPDFKit](https://github.com/PSPDFKit/PSPDFKitSwift/blob/master/Rakefile).

These tasks will later be called from CircleCI during the building/testing stages.

Rakefile
```diff
@@ -0,0 +1,67 @@
...
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program; if not, write to the Free Software Foundation,
+# Inc., 51 Franklin Street, Fifth Floor, Boston MA 02110-1301, USA.
+#
+#
+#
+# ------------------------------------------------------------- Constants ------
+
+SDK_SIM = "iphonesimulator11.3"
```
carolanitz: tests should always run on the lowest supported os version. For CircleCi I guess this would be 10.0 but it would be good if you double check

Rakefile
```diff
@@ -0,0 +1,67 @@
...
+SDK_SIM_DEST = "'platform=iOS Simulator,name=iPhone 7,OS=11.3'"
+
+SCHEME_IOS = "MobileVLCKit"
+PROJECT_IOS = "MobileVLCKit.xcodeproj"
+
+VLC_FLAGS_SIM = "-da x86_64"
```
carolanitz: If you only use SDK_SIM_DEST, SDK_SIM, VLC_FLAGS_SIM in one place there is no need to define them here.

Rakefile
```diff
@@ -0,0 +1,67 @@
...
+# Inc., 51 Franklin Street, Fifth Floor, Boston MA 02110-1301, USA.
+#
+#
+#
+# ------------------------------------------------------------- Constants ------
+
+SDK_SIM = "iphonesimulator11.3"
+SDK_SIM_DEST = "'platform=iOS Simulator,name=iPhone 7,OS=11.3'"
+
+SCHEME_IOS = "MobileVLCKit"
```
carolanitz: right now the testing and build scheme are called the same which is very confusing. Can we rename it to also include the name test?

Rakefile
```diff
@@ -0,0 +1,67 @@
...
+
+desc "Build VLCKit (iOS Simulator)"
+task "build:vlckit:ios" do
+  puts "Building VLCKit (iOS Simulator)"
+
+  plugin_file = "Resources/MobileVLCKit/vlc-plugins-iPhone.h"
+  required_dirs = ["./libvlc/vlc/install-iPhone", "./libvlc/vlc/install-iPhoneSimulator", "./libvlc/vlc/build-iPhoneSimulator"]
+
+  if File.exist?(plugin_file) and dirs_exist?(required_dirs)
+    puts "Found pre-existing build directory. Skipping build"
```
carolanitz: I actually think you don't need to check for this since the compileAndBuild will be already fast if those directories exist

Rakefile
```diff
@@ -0,0 +1,67 @@
...
+VLC_FLAGS_SIM = "-da x86_64"
+
+# ----------------------------------------------------------------- Tasks ------
+
+desc "Build VLCKit (iOS Simulator)"
+task "build:vlckit:ios" do
+  puts "Building VLCKit (iOS Simulator)"
+
+  plugin_file = "Resources/MobileVLCKit/vlc-plugins-iPhone.h"
+  required_dirs = ["./libvlc/vlc/install-iPhone", "./libvlc/vlc/install-iPhoneSimulator", "./libvlc/vlc/build-iPhoneSimulator"]
```
carolanitz: do you really need iPhone when you just build for the Simulator?

Rakefile
Mikanbu: Do we maybe want to add `-v` option as well for more info?

Rakefile
mkchoi212: Hmmm on my machine, it takes about 15 minutes to do a build even when those directories exist 🤔

Rakefile
mkchoi212: 👍

Rakefile
mkchoi212: @carolanitz I took them out as vars because they are going to be used often later on. i.e. when task for code coverage is added on

Rakefile
mkchoi212: So I added `MobileVLCKit-iOSTests` as one of the testable schemes in `Manage Schemes` in Xcode but because of `project.xcworkspace` in `.gitignore`, I can't commit the change 🤔

Rakefile
mkchoi212: Also I'm not 100% how it works yet but I think every time a CI build is triggered, it clones a fresh version of the repo, therefore removing the previous build directories

Rakefile
```diff
@@ -0,0 +1,60 @@
...
+task 'test:ios' do
+  puts 'Running tests for MobileVLCKit'
+  sh "xcodebuild -project #{PROJECT_IOS} -scheme #{SCHEME_IOS} -sdk #{SDK_SIM} -destination #{SDK_SIM_DEST} test"
+end
+
+# ------------------------------------------------------------- Functions ------
+
+def dirs_exist?(directories)
+  directories.each do |dir|
+    return false unless Dir.exist?(dir)
```
carolanitz, line 58: can't you just do this?
```
return Dir.exist?(dir)
```

mkchoi212: Are you talking about replacing the for-loop with `Dir.exist?(dir_1) && Dir.exist?(dir_2)`?

Rakefile
carolanitz: Yes, that's actually true you would need to build a cache yourself. We would need to make sure that we don't use the cache when we add for example a new patch or change something in the compileAndBuildVLCKit.sh

Rakefile
carolanitz: no just replace the ```return false unless Dir.exist?(dir)``` with ```return Dir.exist?(dir)``` but I actually don't know enough about ruby. So I'm genuinely asking because I don't know

Rakefile
mkchoi212: If `compileAndBuildVLCKit.sh` changes, CircleCI will do a rebuild. It's done in `.circle.ci/config.yml` with this 
```
- save_cache:
    key: libvlc-{{ checksum "compileAndBuildVLCKit.sh" }}
```
I'm not sure about adding new patches though cause I haven't really figured out how patches work yet :p

Rakefile
mkchoi212: If you do that, the for-loop would return after only evaluating the first element in the `directories` array. 

`return false unless Dir.exist?(dir)`, returns a `false` if and only if `Dir.exist?(dir)` evaluates to `false`.

If the loop doesn't return, the function implicitly returns a `true`

mkchoi212: I originally had something like 
```ruby
def dirs_exist?(directories)
  directories.each { |dir| 
    if Dir.exist?(x) == false { 
       return false
    }
  }
  return true 
```
but rubocop kept yelling at me until I got to what I have right now :D

carolanitz (approved):

mkchoi212:
Merged with https://github.com/videolan/vlckit/commit/cb34d5608923aca2e5340f3e021720df8afd83f1

mkchoi212:
I just saw that https://github.com/videolan/vlckit/pull/14/commits/23515c7f393641735e4585e1af3d18d7fcc13de6 hasn't been merged yet.

Mikanbu:
Seems that it has with cb34d5608923aca2e5340f3e021720df8afd83f1

PR #16, opened 2018-05-28
Code coverage

mkchoi212:
Code coverage implementation on CircleCI along with pretty output with xcpretty during testing has been implemented.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,6 +12,13 @@ jobs:
     steps:
       - checkout
       - restore_cache:
+          key: v1-gems-{{ checksum "Gemfile.lock" }}
+      - run: bundle check || bundle install --path .bundle
+      - save_cache:
+          key: v1-gems-{{ checksum "Gemfile.lock" }}
+          paths:
+            - .bundle
+      - restore_cache:
           key: libvlc-{{ checksum "compileAndBuildVLCKit.sh" }}
       - run:
           name: Build VLCKit
@@ -27,4 +34,11 @@ jobs:
           name: Run Tests
           command: |
             rake test:ios
+      - run:
+          name: Generate Coverage Reports
+          command: |
+            rake codecov:ios
+      - store_artifacts:
+          path: Tests/Coverage
+          destination: Coverage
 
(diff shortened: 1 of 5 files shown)


mkchoi212:
Dummy CircleCI build with some testing code passed 🎉

You can checkout the static html generated by slather [here](https://6-125473312-gh.circle-artifacts.com/0/Coverage/MobileVLCKitTests/index.html) and the CircleCI build [here](https://circleci.com/gh/videolan/vlckit/6#artifacts/containers/0)

mkchoi212:
Gave up on slather because it had some major bugs that need fixing. And Caro brought up a good point about trying to limit the number of dependencies.

So instead of slather, Apple's own xccov is being used. The [report is rather plain](https://9-125473312-gh.circle-artifacts.com/0/Coverage/MobileVLCKitTests_coverage.txt) but as services like codecov.io start adopting the new format, I think we should look into using their services for making the reports look more comprehensible.

Here's a [dummy CircleCI build](https://circleci.com/gh/videolan/vlckit/9#artifacts/containers/0) with some test code in it.

mkchoi212:
Gave up on slather because it had some [major bugs](https://github.com/SlatherOrg/slather/issues/393) that need fixing. And Caro brought up a good point about trying to limit the number of dependencies.

So instead of slather, Apple's xccov is being used. The [generated report is rather plain](https://9-125473312-gh.circle-artifacts.com/0/Coverage/MobileVLCKitTests_coverage.txt) but as services like [codecov.io](https://codecov.io) start adopting the new format, I think we should look into using their services for making the reports look more comprehensible.

Here's a [dummy CircleCI build](https://circleci.com/gh/videolan/vlckit/9#artifacts/containers/0) with some test code in it.

.circleci/config.yml
```diff
@@ -27,4 +34,11 @@ jobs:
           name: Run Tests
           command: |
             rake test:ios
+      - run:
+          name: Generate Coverage Reports
+          command: |
+            rake codecov:ios
```
carolanitz, line 40: 👍 1:

carolanitz (approved):
looks good to me

carolanitz:
merged with 53dfdb4b4eb2830b24b9456ea4506ef4fd36b4aa

PR #17, opened 2018-06-05
VLCKit (macOS) Tests

mkchoi212:
The following PR needs to be rebased once https://github.com/videolan/vlckit/pull/16 is merged.

In this PR,

- Test scheme for VLCKit has been created
- build / test/ code coverage tasks have been added to the `Rakefile`
- `.circle.yml` has been updated

PR #18, opened 2018-06-07
Setup TVVLCKitTests

mkchoi212:
Commits from https://github.com/videolan/vlckit/pull/17 have been cherry-picked and will need to be rebased on it is merged

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,14 +19,15 @@ jobs:
           paths:
             - .bundle
       - restore_cache:
-          key: libvlc-v2-{{ checksum "compileAndBuildVLCKit.sh" }}
+          key: libvlc-v4-{{ checksum "compileAndBuildVLCKit.sh" }}
       - run:
           name: Build VLCKit
           command: |
             rake build:vlckit:ios
+            rake build:vlckit:tv
             rake build:vlckit:mac
       - save_cache:
-          key: libvlc-v2-{{ checksum "compileAndBuildVLCKit.sh" }}
+          key: libvlc-v4-{{ checksum "compileAndBuildVLCKit.sh" }}
           paths:
             - build
             - libvlc
@@ -35,11 +36,13 @@ jobs:
           name: Run Tests
           command: |
             rake test:ios
+            rake test:tv
             rake test:mac
       - run:
           name: Generate Coverage Reports
           command: |
             rake codecov:ios
+            rake codecov:tv
             rake codecov:mac
       - store_artifacts:
           path: Tests/Coverage
--- a/MobileVLCKit.xcodeproj/project.pbxproj
+++ b/MobileVLCKit.xcodeproj/project.pbxproj
@@ -180,7 +180,17 @@
 		8D826AE2203F3CC500536EF7 /* VLCRendererItem.h in Copy Headers */ = {isa = PBXBuildFile; fileRef = 8D2CE64E203EEA47004BB7F6 /* VLCRendererItem.h */; };
 		AACBBE4A0F95108600F1A2B1 /* Foundation.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = AACBBE490F95108600F1A2B1 /* Foundation.framework */; };
 		CA0301D020ADFD2500452BB8 /* libMobileVLCKit.a in Frameworks */ = {isa = PBXBuildFile; fileRef = D2AAC07E0554694100DB518D /* libMobileVLCKit.a */; };
-		CA8A2F5620B44BD600F5D3EA /* VLCTimeTest.swift in Sources */ = {isa = PBXBuildFile; fileRef = CA8A2F5520B44BD600F5D3EA /* VLCTimeTest.swift */; };
+		CAA03D0C20C8DCDB001764DA /* libTVVLCKit.a in Frameworks */ = {isa = PBXBuildFile; fileRef = 7D1329591BA304D900BE647E /* libTVVLCKit.a */; };
+		CAA03D1720C8DE4D001764DA /* libbz2.tbd in Frameworks */ = {isa = PBXBuildFile; fileRef = CAA03D1620C8DE4D001764DA /* libbz2.tbd */; };
+		CAA03D1920C8DE51001764DA /* libz.tbd in Frameworks */ = {isa = PBXBuildFile; fileRef = CAA03D1820C8DE51001764DA /* libz.tbd */; };
+		CAA03D1B20C8DE58001764DA /* libiconv.tbd in Frameworks */ = {isa = PBXBuildFile; fileRef = CAA03D1A20C8DE58001764DA /* libiconv.tbd */; };
+		CAA03D1D20C8DE5E001764DA /* libc++.tbd in Frameworks */ = {isa = PBXBuildFile; fileRef = CAA03D1C20C8DE5E001764DA /* libc++.tbd */; };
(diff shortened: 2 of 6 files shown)


fkuehne (approved):
Nice work!

.circleci/config.yml
```diff
@@ -19,14 +19,15 @@ jobs:
           paths:
             - .bundle
       - restore_cache:
-          key: libvlc-v2-{{ checksum "compileAndBuildVLCKit.sh" }}
+          key: libvlc-v4-{{ checksum "compileAndBuildVLCKit.sh" }}
```
carolanitz, line 22: I'm curious why do you need to change that?

MobileVLCKit.xcodeproj/project.pbxproj
```diff
@@ -180,7 +180,17 @@
 		8D826AE2203F3CC500536EF7 /* VLCRendererItem.h in Copy Headers */ = {isa = PBXBuildFile; fileRef = 8D2CE64E203EEA47004BB7F6 /* VLCRendererItem.h */; };
 		AACBBE4A0F95108600F1A2B1 /* Foundation.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = AACBBE490F95108600F1A2B1 /* Foundation.framework */; };
 		CA0301D020ADFD2500452BB8 /* libMobileVLCKit.a in Frameworks */ = {isa = PBXBuildFile; fileRef = D2AAC07E0554694100DB518D /* libMobileVLCKit.a */; };
-		CA8A2F5620B44BD600F5D3EA /* VLCTimeTest.swift in Sources */ = {isa = PBXBuildFile; fileRef = CA8A2F5520B44BD600F5D3EA /* VLCTimeTest.swift */; };
+		CAA03D0C20C8DCDB001764DA /* libTVVLCKit.a in Frameworks */ = {isa = PBXBuildFile; fileRef = 7D1329591BA304D900BE647E /* libTVVLCKit.a */; };
+		CAA03D1720C8DE4D001764DA /* libbz2.tbd in Frameworks */ = {isa = PBXBuildFile; fileRef = CAA03D1620C8DE4D001764DA /* libbz2.tbd */; };
+		CAA03D1920C8DE51001764DA /* libz.tbd in Frameworks */ = {isa = PBXBuildFile; fileRef = CAA03D1820C8DE51001764DA /* libz.tbd */; };
+		CAA03D1B20C8DE58001764DA /* libiconv.tbd in Frameworks */ = {isa = PBXBuildFile; fileRef = CAA03D1A20C8DE58001764DA /* libiconv.tbd */; };
```
carolanitz, line 186: why do we need to add all of these frameworks?

MobileVLCKit.xcodeproj/project.pbxproj
```diff
@@ -505,16 +550,25 @@
 				7D89394F1B500DBC008F2B14 /* StaticLibVLC */,
 				7D13295A1BA304D900BE647E /* TVVLCKit */,
 				7DB683801C995D39000C70BE /* DynamicTVVLCKit */,
-				CA0301CC20ADFD2500452BB8 /* MobileVLCKitTests */,
+				CAA9EFFC20D2546000CDBB2C /* Tests */,
 				0867D69AFE84028FC02AAC07 /* Frameworks */,
 				034768DFFF38A50411DB9C8B /* Products */,
+				CAA9EFFB20D24AEE00CDBB2C /* Recovered References */,
```
carolanitz: why do we need that folder?

Rakefile
```diff
@@ -22,18 +22,23 @@
...
 SDK_SIM_DEST_IOS = "'platform=iOS Simulator,name=iPhone 7,OS=11.3'"
+SDK_SIM_DEST_TV = "'platform=tvOS Simulator,name=Apple TV,OS=11.3'"
 SDK_SIM_DEST_MAC = "'platform=OS X,arch=x86_64'"
 
 SCHEME_IOS = 'MobileVLCKitTests'
+SCHEME_TV = 'TVVLCKitTests'
 SCHEME_MAC = 'VLCKitTests'
 
 VLC_FLAGS_IOS = '-dva x86_64'
+VLC_FLAGS_TV = '-ftb'
```
carolanitz: would -stb be enough? we don't need it for a device here, do we?

Rakefile
```diff
@@ -57,6 +62,20 @@ task 'build:vlckit:ios' do
...
+  puts 'Building TVVLCKit'
+
+  plugin_file = 'Resources/MobileVLCKit/vlc-plugins-AppleTV.xcconfig'
+  required_dirs = ['./libvlc/vlc/install-AppleTVSimulator', './libvlc/vlc/build-AppleTVSimulator']
+
+  if File.exist?(plugin_file) && dirs_exist?(required_dirs)
+    puts 'Found pre-existing build directory. Skipping build'
+  else
+    sh "./compileAndBuildVLCKit.sh #{VLC_FLAGS_TV}"
+  end
```
carolanitz, line 76: I am still not sure if we really need to skip the build here.  I see this as a potential source for a wrong test outcome where we use an outdated build directory. We right now don't need to be fast or save time so our number 1 priority should just be an accurate test.

carolanitz (commented):
I just have a couple of questions to understand parts of this PR. I don't want to blindly merge this :) and remarks for thought

.circleci/config.yml
mkchoi212: CircleCI doesn't let you override old caches. With this PR, new directories regarding TVVLCKit must be cached. Also, I went from `v2` to `v4` because I messed up with `v3` :p

Rakefile
mkchoi212: That's true. But if we have the right build directory cached to start off with, I can't really see anything going wrong - since we aren't modifying anything in the build directory. Plus, if anything changes to the way we build VLC, the old cache would be ignored.

I think by caching the build directory, we have one less thing to worry about.

mkchoi212: What we could also do is keep using the cache for now and remove it once all tests are written. This way we don't have to wait for an hour to build & run tests while submitting minor patches/additions.

MobileVLCKit.xcodeproj/project.pbxproj
mkchoi212: We don't :D good catch

MobileVLCKit.xcodeproj/project.pbxproj
mkchoi212: I had the hardest time trying to figure this out but @fkuehne mentioned [here](https://code.videolan.org/videolan/VLCKit/issues/69) that

> For the C++ errors - do you link against libc++ and all the other required libraries? See a dynamic framework target to get an idea.

The other required libraries he was talking about are these libraries. I'm currently writing tests in `MobileVLCKitTests` and I had to include all frameworks from `DynamicMobileVLCKit` as well.

Rakefile
mkchoi212: Seems like using `-stb` fails to compile 🤔https://circleci.com/gh/videolan/vlckit/42

carolanitz: @Mikanbu fixed the underlying issue can we retry with stb here?

carolanitz:
I'd say this is blocked by https://code.videolan.org/videolan/VLCKit/issues/178 or more this is exactly the reason why we need this test :D to catch this!

carolanitz:
this was merged with fc5e423ab3b08ec74822f842ab5d14e8b9b8a145

PR #19, opened 2018-06-19
Create VLCMedia Tests

mkchoi212:
Commits need to be rebased once https://github.com/videolan/vlckit/pull/18 is merged.


Sources/VLCMedia.m
```diff
@@ -188,10 +188,7 @@ + (NSString *)codecNameForFourCC:(uint32_t)fourcc trackType:(NSString *)trackTyp
         track_type = libvlc_track_text;
 
     const char *ret = libvlc_media_get_codec_description(track_type, fourcc);
-    if (ret)
-        return [NSString stringWithUTF8String:ret];
-
-    return @"";
+    return [NSString stringWithUTF8String:ret];
```
carolanitz: this will crash with ret = nil. This was not redundant line

fkuehne (approved):

fkuehne:
After removing the commit NACKed by Caro, this is good to go for me and I like it :)

Sources/VLCMedia.m
mkchoi212: I tracked it down and seems like `*ret = libvlc_media_get_codec_description(...)` calls `vlc_fourcc_GetDescription`, which has the following return statement

```c
return LookupCat(fourcc, &ret, cat) ? ret : "";
```

Doesn't this mean that `vlc_fourcc_GetDescription` will never return nil?

mkchoi212:
Ping 🔔

carolanitz:
merged with faf2a0d56bc2afbd1d6e954c054df9c181404fe4

PR #20, opened 2018-06-22
buildscript: Fix tvOS build

Mikanbu:
For example, if we choose the compile only aarch64(with -a), it would go into the simulator part and fail since arm64 is not a simulator platform.


carolanitz:
merged with 2a09095e8185aff0797ad5515aa63e31ca49d642

PR #21, opened 2018-06-28
build: Fix simulators build (closes #178)

Mikanbu:
In `build_universal_static_lib()` the architecture was set by default to arm64 which led to `No such file or directory` for simulators(`-s`).

### DIFF
--- a/compileAndBuildVLCKit.sh
+++ b/compileAndBuildVLCKit.sh
@@ -755,7 +755,10 @@ build_universal_static_lib() {
     arch="aarch64"
     if [ "$FARCH" != "all" ];then
         arch="$FARCH"
+    elif [ "$BUILD_SIMULATOR" = "yes" ]; then
+        arch="x86_64"
     fi
+
     actual_arch=`get_actual_arch $arch`
 
     if [ -d ${VLCROOT}/install-"$OSSTYLE"OS ];then
--- a/compileAndBuildVLCKit.sh
+++ b/compileAndBuildVLCKit.sh
@@ -800,38 +800,37 @@ build_universal_static_lib() {
 
     if [ -d ${VLCROOT}/install-"$OSSTYLE"Simulator ];then
         spushd ${VLCROOT}/install-"$OSSTYLE"Simulator
-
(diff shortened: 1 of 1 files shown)


compileAndBuildVLCKit.sh
```diff
@@ -755,7 +755,10 @@ build_universal_static_lib() {
     arch="aarch64"
     if [ "$FARCH" != "all" ];then
         arch="$FARCH"
+    elif [ "$BUILD_SIMULATOR" = "yes" ]; then
+        arch="x86_64"
```
fkuehne, line 759: This is inaccurate. There are i686 builds for iOS Simulator.

compileAndBuildVLCKit.sh
```diff
@@ -797,38 +800,37 @@ build_universal_static_lib() {
 
     if [ -d ${VLCROOT}/install-"$OSSTYLE"Simulator ];then
         spushd ${VLCROOT}/install-"$OSSTYLE"Simulator
-
-        if (is_simulator_arch $arch);then
-            echo "SIMU OS: $arch"
-            spushd $actual_arch/lib/vlc/plugins
-            for i in `ls *.a`
+            for i in `ls .`
```
fkuehne, line 803: Why drop the .a check?

compileAndBuildVLCKit.sh
Mikanbu: Isn't it the same for non-simulator build then since there are not only aarch64?

compileAndBuildVLCKit.sh
Mikanbu: It shouldn't have, it's just the block moving up.

compileAndBuildVLCKit.sh
tguillem: @fkuehne this arch variable is used to get the list of modules. So, it will works for i686 builds too (like it works for armv7 build when aarch is aarch64).

So, LGTM.

carolanitz (approved):

fkuehne (approved):

PR #23, opened 2018-07-05
Tests: Add video samples

mkchoi212:
4 sample videos have been downloaded and added.
Each video is about 3-5 seconds for the total size of **1.5 MB.**

All videos have been downloaded from https://vimeo.com/user32335610 and are under the  [Creative Commons Public Domain Dedication license](https://creativecommons.org/publicdomain/zero/1.0/).

After downloading the clips, I shortened the videos and changed the file formats to diversify the sample.

### DIFF
--- a/Tests/Helper/Video.swift
+++ b/Tests/Helper/Video.swift
@@ -1,0 +1,89 @@
+/*****************************************************************************
+ * Video.swift
+ *****************************************************************************
+ * Copyright (C) 2018 Mike JS. Choi
+ * Copyright (C) 2018 VLC authors and VideoLAN
+ * $Id$
+ *
+ * Authors: Mike JS. Choi <mkchoi212 # icloud.com>
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation; either version 2.1 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston MA 02110-1301, USA.
+ *****************************************************************************/
+
+import XCTest
+
+struct Video {
+    let name: String
+    let type: String
+    let meta: [String:String]
+    
+    #if os(OSX)
+        let bundle = Bundle(identifier: "org.videolan.VLCKitTests")!
+    #elseif os(iOS)
+        let bundle = Bundle(identifier: "org.videolan.MobileVLCKitTests")!
+    #elseif os(tvOS)
+        let bundle = Bundle(identifier: "org.videolan.TVVLCKitTests")!
+    #endif
+    
+    static let standards = [Video.test1, Video.test2, Video.test3, Video.test4]
+    
+    static let test1 = Video(
+        name: "bird",
+        type: "m4v",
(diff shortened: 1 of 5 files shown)


fkuehne:
Is it possible to store those videos online (we can provide storage space) and download them on-the-fly so we don't need to add so much binary data to the git?

mkchoi212:
Good idea. If you could upload them to a server and provide me a link to them, that would be great!

Tests/Assets/download_videos.py
```diff
@@ -0,0 +1,40 @@
...
+download_videos.py
+
+Download 4 videos used during VLCKit tests
+'''
+import os
+import urllib
+
+# URL and desired file names of four videos to be downloaded
+VIDEOS = [
+    ('https://www.sample-videos.com/video/mp4/240/big_buck_bunny_240p_1mb.mp4', 'bird.m4v'),
```
mkchoi212: These links must be updated :D

Tests/Helper/Video.swift
```diff
@@ -0,0 +1,81 @@
...
+        type: "mov",
+        meta: ["description": "Tanning sea lions", "genre": "Nature", "title": "sea_lions.mov"]
+    )
+    static let invalid = Video(
+        name: "invalid",
+        type: "foo",
+        meta: [:]
+    )
+    
+    let bundle = Bundle(for: VLCTimeTest.self)
```
mkchoi212: Is there a better way to load the testing bundle here?

carolanitz:
if we upload it somewhere we have the same issues as in the mainrepo which is tests intermittedly failing because of slow connection or failing connection and if the tests are not reliable they're useless

fkuehne:
Then we should have a git submodule for the files. Adding 6MB of binary blobs to the git just for testing will increase download time and used disk space for everyone, so it would be nicer to optionalize this.

carolanitz:
Can we look into best practices for this usecase? We're not the only ones who deal with this problem so i'd like to know how other projects handle this case

jbkempf:
Git-LFS is an option, but a pain to synchronize between repos.

The other is a local cache from distant files, like the contrib tarballs (SHA-512) and is probably the simplest.

carolanitz:
Git-LFS is for large files, we're talking 1,5 MB here and isn't absolutely common to check in files for testing into a repo ?

carolanitz:
btw don't get me wrong, this is absolutely a valid point that we might unnecessarily enlarge our repository here.  I could've sworn though that checking test data into your repository is common practice so I didn't even second guess it.

Tests/Assets/download_videos.py
```diff
@@ -0,0 +1,50 @@
...
+download_videos.py
+
+Download videos used during VLCKit tests
+'''
+import os
+import urllib
+
+# URL and desired file names of videos to be downloaded
+VIDEOS = [
+    ('https://www.sample-videos.com/video/mp4/240/big_buck_bunny_240p_1mb.mp4', 'bird.m4v'),
```
mkchoi212: These links will be replaced by links for code.v.o once available

carolanitz:
could you rebase this on master ? or does circleci failing because of something else ?

mkchoi212:
The PR for TestAssets hasn’t been merged yet :(

I’ll ping thresh

mkchoi212:
From [Twitter](https://twitter.com/Bananamlkshake2/status/1022417680928653312), it seems like loading the videos from the test bundle after being downloaded is the best way to go; which is how this PR is setup.

Pros are 

- **Tests can't start** if the assets can't be downloaded from code.v.o/TestAssets
   - This is good because if we don't load the videos from the bundle, the tests will start and **fail** with messages like "*can't open file x*"
- Buying into Xcode's infra

Cons are

- Minor complexity during initial test run
- Additional files added to `.pbxproj`

carolanitz:
merged with commit aa98dd30e705f7243f41db174914827a3a1e2334 &&
commit 1be28e18561376c163cb76b69fa9585122cc6d98 &&
commit 458b355615988105e1350c6ba5029aa106645052

PR #28, opened 2018-07-17
VLCDialogProvider: Adjust DialogProvider to not show errordialogs (cl…

carolanitz:
…oses #184)

we might want to still show them in debug mode. I'm not sure if removing it entirely is okay here.


fkuehne (changes requested):
I oppose this change. Especially with the custom dialog renderer protocol, it is completely optional to the client app to implement support for error panels or not.
If the author of the client app decides not to show those dialogs or only in debug mode, this is already possible. However, displaying errors in debug mode or in production will be impossible for apps using this protocol if you remove this API.

carolanitz:
In that case we won't have to touch VLCKit but VLC for iOS

PR #30, opened 2018-07-20
VLCLibrary: debugLogging status (closes #186)

mkchoi212:
Fixes the bugs

- `debugLoggingStatus` can not be set
- `debugLoggingLevel` can exceed the set range of 0-4

### DIFF
--- a/Headers/Public/VLCLibrary.h
+++ b/Headers/Public/VLCLibrary.h
@@ -30,86 +30,79 @@
 @class VLCAudio;
 
 /**
- * The VLCLibrary is the base library of the VLCKit.framework.  This object provides a shared instance that exposes the
+ * The VLCLibrary is the base library of VLCKit.framework. This object provides a shared instance that exposes the
  * internal functionalities of libvlc and libvlc-control. The VLCLibrary object is instantiated automatically when
- * VLCKit.framework is loaded into memory.  Also, it is automatically destroyed when the VLCKit.framework is unloaded
+ * VLCKit.framework is loaded into memory.  Also, it is automatically destroyed when VLCKit.framework is unloaded
  * from memory.
  *
- * Currently, the framework does not support multiple instances of VLCLibrary.  Furthermore, you cannot destroy any
- * instiantiation of VLCLibrary, as previously noted, this is done automatically by the dynamic link loader.
+ * Currently, the framework does not support multiple instances of VLCLibrary.
+ * Furthermore, you cannot destroy any instance of VLCLibrary; this is done automatically by the dynamic link loader.
  */
 @interface VLCLibrary : NSObject
 
-/* Factories */
 /**
- * Returns the library's shared instance.
- * \return The library's shared instance.
+ * Returns the library's shared instance
+ * \return The library's shared instance
  */
 + (VLCLibrary *)sharedLibrary;
 
 /**
- * returns an individual instance which can be customized with options
- * \param options a NSArray with NSString instance containing the options
+ * Returns an individual instance which can be customized with options
+ * \param options NSArray with NSString instance containing the options
  * \return the individual library instance
  */
- - (instancetype)initWithOptions:(NSArray*)options;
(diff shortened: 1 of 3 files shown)


Sources/VLCLibrary.m
```diff
@@ -143,13 +143,27 @@ - (void)setDebugLogging:(BOOL)debugLogging
...
 
+- (void)setDebugLoggingLevel:(int)debugLoggingLevel
+{
+    if (debugLoggingLevel >= 0 && debugLoggingLevel <= 4) {
+        _debugLoggingLevel = debugLoggingLevel;
+    } else {
+        VKLog(@"Invalid debugLoggingLevel of %d provided", debugLoggingLevel);
+        VKLog(@"Please provide a valid debugLoggingLevel between 0 and 4");
+        VKLog(@"Defaulting debugLoggingLevel to 4 (just errors)");
+        _debugLoggingLevel = 4;
```
mkchoi212: Decided to set a default value for the logging level because I think an invalid debug logging level isn't worth causing a crash.

fkuehne: The scale is the other way around - 0 is just errors, whereas 4 is "everything VLC ever wanted to tell you about anything". So the default should be 0.
However, there is something wrong if there is a crash when the user provides 5. Are you sure that this is happening?

mkchoi212: Gotcha! Also, it doesn't crash. What I meant was that I initially thought about doing `fatalError("Invalid debugging level x")` but just decided to warn the user about it.

carolanitz: What do you think about  just capping the value between 0 and 4 meaning <0 to 0 >4 to 4 and if it's outside those bounds log the message ?

Sources/VLCLibrary.m
```diff
@@ -143,13 +143,27 @@ - (void)setDebugLogging:(BOOL)debugLogging
...
     if (debugLogging) {
         libvlc_log_set(_instance, HandleMessage, (__bridge void *)(self));
     } else {
         libvlc_log_unset(_instance);
     }
 }
 
+- (void)setDebugLoggingLevel:(int)debugLoggingLevel
+{
+    if (debugLoggingLevel < 0 && debugLoggingLevel > 4) {
```
mkchoi212: @carolanitz something like this?? I’m mildly confused :D

Headers/Public/VLCLibrary.h
```diff
@@ -30,86 +30,79 @@
...
+ * The VLCLibrary is the base library of VLCKit.framework. This object provides a shared instance that exposes the
  * internal functionalities of libvlc and libvlc-control. The VLCLibrary object is instantiated automatically when
- * VLCKit.framework is loaded into memory.  Also, it is automatically destroyed when the VLCKit.framework is unloaded
+ * VLCKit.framework is loaded into memory.  Also, it is automatically destroyed when VLCKit.framework is unloaded
  * from memory.
  *
- * Currently, the framework does not support multiple instances of VLCLibrary.  Furthermore, you cannot destroy any
- * instiantiation of VLCLibrary, as previously noted, this is done automatically by the dynamic link loader.
+ * Currently, the framework __does not__ support multiple instances of VLCLibrary.
+ * Furthermore, you __cannot__ destroy any instance of VLCLibrary; this is done automatically by the dynamic link loader.
```
fkuehne: I'm not sure if this underling syntax is permitted in doxygen. Can you verify this?

fkuehne (commented):
One nit-pick, rest looks good now! :)

Headers/Public/VLCLibrary.h
mkchoi212: Ah I had no idea this was run through doxygen. And turns out, doxygen only supports `__` syntax for markdown.

mkchoi212:
ping 🔔

carolanitz:
merged

PR #31, opened 2018-07-22
Added initializing VLCMedia from NSInputStream (using callbacks)

amosavian:
This PR adds `[VLCMedia initWithStream:]` initializer, also introduces `VLCHTTPInputStream` class to utilize `NSURLSession` for streaming http remote media.

`VLCHTTPInputStream` class can be used to deprecate VLCMedia initializing from http url as it uses `NSURLSession`, which can handle cellular data and multipath tcp. This class implements an internal cache in order to maintain smooth playing experience.

Later we can implement an stream using [sahlberg/libsmb2](https://github.com/sahlberg/libsmb2) to stream from smb2 servers and remove libdsm eventually.

[Discussion here](https://code.videolan.org/videolan/VLCKit/issues/162#note_15643).

### DIFF
--- a/Sources/VLCMedia.m
+++ b/Sources/VLCMedia.m
@@ -65,6 +65,54 @@ NSString *const VLCMetaInformationDiscNumber     = @"discNumber";
 NSString *const VLCMediaMetaChanged              = @"VLCMediaMetaChanged";
 
 /******************************************************************************
+ * VLC callbacks for streaming.
+ */
+int open_cb(void *opaque, void **datap, uint64_t *sizep) {
+    NSInputStream *stream = (__bridge NSInputStream *)(opaque);
+    
+    // Once a stream is closed, it cannot be reopened.
+    if (stream && stream.streamStatus == NSStreamStatusNotOpen) {
+        [stream open];
+        *datap = opaque;
+        *sizep = UINT64_MAX;
+        return 0;
+    }
+    return stream.streamStatus == NSStreamStatusOpen ? 0 : -1;
+}
+
+ssize_t read_cb(void *opaque, unsigned char *buf, size_t len) {
+    NSInputStream *stream = (__bridge NSInputStream *)(opaque);
+    if (!stream) {
+        return -1;
+    }
+    
+    return [stream read:buf maxLength:len];
+}
+
+int seek_cb(void *opaque, uint64_t offset) {
+    NSInputStream *stream = (__bridge NSInputStream *)(opaque);
+    if (!stream) {
+        return -1;
+    }
+    
+    /*
+     By default, NSStream instances that are not file-based are non-seekable, one-way streams (although custom seekable subclasses are possible).
+     Once the data has been provided or consumed, the data cannot be retrieved from the stream.
+     
+     However, you may want a peer subclass to NSInputStream whose instances are capable of seeking through a stream.
+     */
+    return [stream setProperty:@(offset) forKey:NSStreamFileCurrentOffsetKey] ? 0 : -1;
+}
+
+void close_cb(void *opaque) {
+    NSInputStream *stream = (__bridge NSInputStream *)(opaque);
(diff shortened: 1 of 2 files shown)


jbkempf:
I absolutely disagree with re-implementing HTTP outside of libvlc-core.

amosavian:
@jbkempf I don't know which library libvlc uses for networking but BSD sockets have few issues in iOS particularly when using cellular data. Apple describes the advantages in [this keynote](https://developer.apple.com/videos/play/wwdc2018/715/). `NSURLSession` has few advantages in multipath tcp and cellular connection management over BSD sockets. Indeed BSD sockets and CFStream bounded to sockets are **deprecated** in iOS 12.

Another option is to provide both options.

By the way if you think we must remove `VLCHTTPInputStream`, I will remove it.

jbkempf:
Sorry, but you cannot replace the HTTP core of VLC with another one and hope it will work out: It won't.

It will break icecast, adaptive, HLS, DASH, Chomecast and so on...

Also, there is no way iOS removes all BSD sockets. And no, because the latest marketing from Apple Dev Department says it's cool, does not mean it's amazing (remember GCD that would replace pthread?)

amosavian:
I forgot to say, iOS 12 and macOS 10.14 introduce `Network.framework` that is much faster when streaming video either using udp to tcp. This change will allow user to implement faster real-time streaming (e.g. rtsp protocol) by wrapping, and use VLCKit to decode video format.

carolanitz:
@amosavian We definitely have to look into the Network.framework but as @jbkempf says we can't replace the HTTP core in iOS. There are a lot of other dependencies that might break and we don't have the manpower to maintain a separate iOS implementation even if it is better.

amosavian:
iOS won't remove BSD socket neither they are marked explicitly, but if you see video, Apple's engineer describe advantage and says BSD sockets are considered deprecated in iOS 12 onward implicitly due to problems it has.

Also I don't know if libvlc-core does supports http/2 or not. But URLSession provides http/2 and brotli compression for free. Why not provide an option to user to use native Apple's solution? New users can take advantage while established libraries still can use libvlc's http. **I don't suggest to remove HTTP core from libvlc, but I think it's nice to have an alternative if developer desires.**

I think you misunderstood me. I never said to remove libvlc-core implementation. I know libvlc is a cross-platform library. I focus on VLCKit which wraps libvlc for Apple's platforms.

@carolanitz Network.framework does provide plain C api. `VLCHTTPInputStream` is a thin wrapper and does not need considerable maintenance. Furthermore, other developers can use this class as a base to develop their own NSInputStream subclass compatible with vlckit.

amosavian:
@jbkempf @carolanitz I've removed `VLCHTTPInputStream` class impl completely. Now we have just `[VLCMedia initWithStream]` discussed before. In case you wish I can readd it as a separate PR.

jbkempf:
> I don't suggest to remove HTTP core from libvlc, but I think it's nice to have an alternative if developer desires

Sure, but then he should do that in his app. Not in VLCKit.

Sources/VLCMedia.m
```diff
@@ -64,6 +64,54 @@
...
+    NSInputStream *stream = (__bridge NSInputStream *)(opaque);
+    if (!stream) {
+        return -1;
+    }
+    
+    return [stream read:buf maxLength:len];
+}
+
+int seek_cb(void *opaque, uint64_t offset) {
+    NSInputStream *stream = (__bridge NSInputStream *)(opaque);
```
robUx4, line 93: If this is not seakable, just pass NULL instead of seek_cb

amosavian: There is no way to check if stream is seekable or not before opening the stream.

carolanitz:
Thank you very much for the PR merged with 29aa853387bc8924b084e7bcd6da310bed150822 :)

PR #32, opened 2018-07-26
Tests: VLCLibrary

mkchoi212:
Needs to be rebased once #30 is merged

### DIFF
--- a/Headers/Public/VLCLibrary.h
+++ b/Headers/Public/VLCLibrary.h
@@ -23,11 +23,6 @@
  *****************************************************************************/
 
 #import <Foundation/Foundation.h>
-#import "VLCAudio.h"
-#import "VLCMediaList.h"
-#import "VLCMedia.h"
-
-@class VLCAudio;
 
 /**
  * The VLCLibrary is the base library of VLCKit.framework. This object provides a shared instance that exposes the
--- a/Tests/Sources/VLCLibraryTest.swift
+++ b/Tests/Sources/VLCLibraryTest.swift
@@ -1,0 +1,122 @@
+/*****************************************************************************
+ * VLCLibraryTest.swift
+ *****************************************************************************
+ * Copyright (C) 2018 VLC authors and VideoLAN
+ * $Id$
+ *
+ * Authors: Mike JS. Choi <mkchoi212 # icloud.com>
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation; either version 2.1 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston MA 02110-1301, USA.
+ *****************************************************************************/
+
+import XCTest
+
+class VLCLibraryTest: XCTestCase {
+    
+    let paramKey = "VLCParams"
+    
+    override func setUp() {
+        super.setUp()
+        UserDefaults.standard.removeObject(forKey: paramKey)
+    }
+
+    func testSharedLibrary() {
+        XCTAssertNotNil(VLCLibrary.shared())
+        XCTAssertNotNil(VLCLibrary.shared().instance)
+    }
+    
+    func testInitWithOptions() throws {
+        let customLibrary1 = try XCTAssertNotNilAndUnwrap(VLCLibrary(options: ["--verbose=1", "--avi-index=1"]))
+        assertDefaultParameters()
+        XCTAssertNotNil(customLibrary1.instance)
+        
+        UserDefaults.standard.removeObject(forKey: paramKey)
+        
+        let customLibrary2 = try XCTAssertNotNilAndUnwrap(VLCLibrary(options: []))
+        assertDefaultParameters()
+        XCTAssertNotNil(customLibrary2.instance)
+    }
+    
+    func testInit() {
+        let library = VLCLibrary()
+        XCTAssertNotNil(library.instance)
+        assertDefaultParameters()
+    }
+    
+    func testDebugLoggingLevel() throws {
+        let library = try XCTAssertNotNilAndUnwrap(VLCLibrary.shared())
+        XCTAssertEqual(library.debugLoggingLevel, 0)
+        
+        let tests: [(input: Int32, expected: Int32)] = [
+            (0, 0),
+            (3, 3),
+            (4, 4),
+            (100, 0),
+            (-10, 0)
+        ]
+        
+        for (input, expected) in tests {
+            library.debugLoggingLevel = input
+            XCTAssertEqual(library.debugLoggingLevel, expected)
+        }
+    }
+    
+    func testDebugLogging() throws {
+        let library = try XCTAssertNotNilAndUnwrap(VLCLibrary.shared())
+        XCTAssertFalse(library.debugLogging)
+        
+        library.debugLogging = true
+        XCTAssertTrue(library.debugLogging)
+        
+        library.debugLogging = false
+        XCTAssertFalse(library.debugLogging)
+    }
+    
+    func testLibraryDescription() throws {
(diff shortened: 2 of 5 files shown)


Headers/Public/VLCLibrary.h
```diff
@@ -23,93 +23,81 @@
...
 
 /**
- * sets the application name and HTTP User Agend
+ * Sets the application name and HTTP User Agent
  * libvlc will pass it to servers when required by protocol
- * \param readableName human-readable application name, e.g. "FooBar player 1.2.3"
+ * \param readableName Human-readable application name, e.g. "FooBar player 1.2.3"
  * \param userAgent HTTP User Agent, e.g. "FooBar/1.2.3 Python/2.6.0"
  */
 - (void)setHumanReadableName:(NSString *)readableName withHTTPUserAgent:(NSString *)userAgent;
```
mkchoi212: Not entirely sure how this can be tested...

Headers/Public/VLCLibrary.h
```diff
@@ -23,93 +23,81 @@
...
- * Sets some meta-information about the application.
- *
+ * Sets meta-information about the application
  * \param identifier Java-style application identifier, e.g. "com.acme.foobar"
- * \param version application version numbers, e.g. "1.2.3"
- * \param icon application icon name, e.g. "foobar"
+ * \param version Application version numbers, e.g. "1.2.3"
+ * \param icon Application icon name, e.g. "foobar"
  */
 - (void)setApplicationIdentifier:(NSString *)identifier withVersion:(NSString *)version andApplicationIconName:(NSString *)icon;
```
mkchoi212: Same thing here

Tests/Sources/VLCLibraryTest.swift
```diff
@@ -0,0 +1,119 @@
...
+    func testInitWithOptions() throws {
+        let tests = [
+            ["--no-video-title-show", "--verbose=4"],
+            []
+        ]
+        
+        for option in tests {
+            let library = try XCTAssertNotNilAndUnwrap(VLCLibrary(options: option))
+            XCTAssertNotNil(library.instance)
+            assertDefaultParameters()
```
carolanitz: Wouldn't it be easier to understand and recognize why a test fails if you don't have a for loop and dict but write the params to test inline?
I know it's less code but right now I have to take some time to understand what line corresponds to a failing test if it would fail here. I'm also a bit unsure why we need to assertDefaultParams for : ["--no-video-title-show", "--verbose=4"]. If both of these are defaults it probably makes sense to test it as well with options that are not the default

Tests/Sources/VLCLibraryTest.swift
```diff
@@ -0,0 +1,119 @@
...
+    func testDebugLoggingLevel() throws {
+        let library = try XCTAssertNotNilAndUnwrap(VLCLibrary.shared())
+        XCTAssertEqual(library.debugLoggingLevel, 0)
+        
+        let tests: [(input: Int32, expected: Int32)] = [
+            (0, 0),
+            (3, 3),
+            (4, 4),
+            (100, 0),
+            (-10, 0)
```
carolanitz, line 67: 👍

Tests/Sources/VLCLibraryTest.swift
```diff
@@ -0,0 +1,119 @@
...
+        library.debugLogging = false
+        XCTAssertFalse(library.debugLogging)
+    }
+    
+    func testLibraryDescription() throws {
+        let library = try XCTAssertNotNilAndUnwrap(VLCLibrary.shared())
+        
+        XCTAssert(library.version.count > 1)
+        XCTAssert(library.compiler.count > 1)
+        XCTAssert(library.changeset.count > 1)
```
carolanitz: What you're testing for here is a bit unclear for me. Those are strings but the way the test is written it looks like you're testing an array.  What do you think about testing for a not empty string and providing a message like "compiler should hold the string of the libvlccompiler, e.g. "gcc version 4.2.3 (Ubuntu 4.2.3-2ubuntu6)"? That way if the test fails we know roughly what we expected

Tests/Sources/VLCLibraryTest.swift
```diff
@@ -0,0 +1,119 @@
...
+            "--vout=macosx",
+            "--text-renderer=freetype",
+            "--extraintf=macosx_dialog_provider",
+            "--audio-resampler=soxr"
+        ]
+        
+        let defaultParams = UserDefaults.standard.object(forKey: paramKey)
+        
+        #if os(macOS)
+            XCTAssertEqual(defaultParams as? [String], expected)
```
carolanitz, line 117: When is defaultParams ever nil? Does it make sense to make this also not optional?

Tests/Sources/VLCLibraryTest.swift
```diff
@@ -0,0 +1,119 @@
...
+    
+    func testDebugLogging() throws {
+        let library = try XCTAssertNotNilAndUnwrap(VLCLibrary.shared())
+        XCTAssertFalse(library.debugLogging)
+        
+        library.debugLogging = true
+        XCTAssertTrue(library.debugLogging)
+        
+        library.debugLogging = false
+        XCTAssertFalse(library.debugLogging)
```
carolanitz, line 84: :+1:

Tests/Sources/VLCLibraryTest.swift
mkchoi212: I thought about doing 
```swift
XCTAssertEqual(library.compiler, "gcc version 4.2.3 (Ubuntu 4.2.3-2ubuntu6))"
```
but every time the library gets an update, you would have to update the tests as well or else, they would fail. Is this desirable?

Tests/Sources/VLCLibraryTest.swift
mkchoi212: `defaultParams` is set in `VLCLibrary` during an `init`. But the user could override it by using the same key by mistake since the parameters are stored in `UserDefaults.standard`

Tests/Sources/VLCLibraryTest.swift
mkchoi212: There is no way to actually see if the custom options were successfully saved because the options are saved via 

```objc
_instance = libvlc_new(count, lib_vlc_params);
```

where `_instance` is a `void *`.

So I just ended up making two dummy options and checking if the library was initialized successfully.

mkchoi212: Maybe I could init a mock `libvlc_new(count, lib_vlc_params)` and compare the two instances?? Is there a `libvlc_compare_library` method?

carolanitz: I don't think there is but you can cast or bridge the void pointer into the correct object :)

Tests/Sources/VLCLibraryTest.swift
carolanitz: I was thinking more along the lines of XCTAssertFalse (library.compiler. isEmpty, "compiler is empty but it should hold a string of the form e.g. "gcc version 4.2.3 (Ubuntu 4.2.3-2ubuntu6)")

Tests/Sources/VLCLibraryTest.swift
carolanitz: I'm still confused a bit by this test :D  so I overread the nil two lines down for iOS and I thought this test makes sure that the defaultParams are always not nil and set?

mkchoi212: `defaultParams` is nill for iOS and tvOS because `VLCLibrary.m` **DOES NOT** set the default parameters in the UserDefaults `#if TARGET_OS_IPHONE`. It does save the default params in UserDefaults otherwise. https://github.com/videolan/vlckit/blob/ba9d2cd7e5099c480df0e7e3a1376508f9243f1b/Sources/VLCLibrary.m#L101

Tests/Sources/VLCLibraryTest.swift
```diff
@@ -0,0 +1,122 @@
...
+    func testLibraryDescription() throws {
+        let library = try XCTAssertNotNilAndUnwrap(VLCLibrary.shared())
+        
+        let warn: (String) -> (String) = { desirable in
+            return "Should hold a string of form \"\(desirable)\""
+        }
+        
+        XCTAssertFalse(library.version.isEmpty, warn("3.0.4 Vetinari"))
+        XCTAssertFalse(library.compiler.isEmpty, warn("InstalledDir: /Applications/Xcode.app/..."))
+        XCTAssertFalse(library.changeset.isEmpty, warn("3.0.3-1-108-g7039639e6b"))
```
carolanitz, line 96: 👍

carolanitz (approved):

carolanitz:
This was merged with 7734b081d1445ce8dc7468d666125beb7c137814 & 55daaa9ac47843fb6a41289d09ecadb612caf77f

PR #34, opened 2018-08-01
VLCMediaList: removeMediaAt (closes #182)

mkchoi212:
Reference https://code.videolan.org/videolan/VLCKit/issues/182

### DIFF
--- a/Headers/Public/VLCMediaList.h
+++ b/Headers/Public/VLCMediaList.h
@@ -104,12 +104,14 @@ extern NSString *const VLCMediaListItemDeleted;
 - (void)insertMedia:(VLCMedia *)media atIndex:(NSUInteger)index;
 
 /**
- * remove a media from a given position
+ * remove media at position index and return true if the operation was successful.
+ * An unsuccessful operation occurs when the index is greater than the medialists count
  *
  * \param index the index of the media to remove
+ * \return boolean result of the removal operation
  * \note this function silently fails if the list is read-only
  */
-- (void)removeMediaAtIndex:(NSUInteger)index;
+- (BOOL)removeMediaAtIndex:(NSUInteger)index;
 
 /**
  * retrieve a media from a given position
--- a/Headers/Public/VLCMediaList.h
+++ b/Headers/Public/VLCMediaList.h
@@ -123,9 +123,10 @@ extern NSString *const VLCMediaListItemDeleted;
  * retrieve the position of a media item
  *
  * \param media the media object to search for
- * \return the index position of the media in the list or -1 if not found
+ * \return The lowest index of the provided media in the list
+ * If media does not exist in the list, returns NSNotFound.
  */
-- (NSInteger)indexOfMedia:(VLCMedia *)media;
+- (NSUInteger)indexOfMedia:(VLCMedia *)media;
 
 /* Properties */
 /**
--- a/Sources/VLCMedia.m
+++ b/Sources/VLCMedia.m
@@ -300,6 +300,12 @@ static void HandleMediaParsedChanged(const libvlc_event_t * event, void * self)
     return p_md == [media libVLCMediaDescriptor] ? NSOrderedSame : NSOrderedAscending;
 }
 
+- (BOOL)isEqual:(id)other
+{
+    return ([other isKindOfClass: [VLCMedia class]] &&
+            [other libVLCMediaDescriptor] == p_md);
+}
+
 - (VLCTime *)length
 {
     if (!_length) {
--- a/Sources/VLCMediaList.m
+++ b/Sources/VLCMediaList.m
@@ -69,7 +69,8 @@ static void HandleMediaListItemDeleted( const libvlc_event_t * event, void * use
         id self = (__bridge id)(user_data);
         [[VLCEventManager sharedManager] callOnMainThreadObject:self
                                                      withMethod:@selector(mediaListItemRemoved:)
-                                           withArgumentAsObject:@(event->u.media_list_item_deleted.index)];
+                                           withArgumentAsObject:@[@{@"media": [VLCMedia mediaWithLibVLCMediaDescriptor:event->u.media_list_item_deleted.item],
+                                                                    @"index": @(event->u.media_list_item_deleted.index)}]];
     }
 }
 
@@ -163,17 +164,22 @@ static void HandleMediaListItemDeleted( const libvlc_event_t * event, void * use
     libvlc_media_list_insert_media(p_mlist, [media libVLCMediaDescriptor], (int)index);
 }
 
-- (void)removeMediaAtIndex:(NSUInteger)index
+- (BOOL)removeMediaAtIndex:(NSUInteger)index
 {
+    __block BOOL ok = YES;
+    
     dispatch_sync(_serialMediaObjectsQueue, ^{
-        if (index >= [_mediaObjects count])
+        // Remove from cached Media
+        if (index >= [_mediaObjects count]) {
+            ok = NO;
             return;
-        //remove from cached Media
+        }
         [_mediaObjects removeObjectAtIndex:index];
     });
 
-    // Remove it from libvlc's medialist
+    // Remove from libvlc's medialist
     libvlc_media_list_remove_index(p_mlist, (int)index);
+    return ok;
 }
 
 - (VLCMedia *)mediaAtIndex:(NSUInteger)index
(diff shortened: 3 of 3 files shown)


Sources/VLCMediaList.m
```diff
@@ -163,17 +164,18 @@ - (void)insertMedia:(VLCMedia *)media atIndex: (NSUInteger)index
     libvlc_media_list_insert_media(p_mlist, [media libVLCMediaDescriptor], (int)index);
 }
 
-- (void)removeMediaAtIndex:(NSUInteger)index
+- (VLCMedia *)removeMediaAtIndex:(NSUInteger)index
 {
+    VLCMedia *target = [_mediaObjects objectAtIndex:index];
```
mkchoi212: `objectAtIndex` checks if the provided index is in range

Headers/Public/VLCMediaList.h
```diff
@@ -104,12 +104,13 @@ extern NSString *const VLCMediaListItemDeleted;
...
- * remove a media from a given position
+ * remove and return a media from a given position
  *
- * \param index the index of the media to remove
+ * \param index the index of the media to remove. index must be a valid index of the array.
+ * \return media at the given index
  * \note this function silently fails if the list is read-only
  */
-- (void)removeMediaAtIndex:(NSUInteger)index;
+- (VLCMedia *)removeMediaAtIndex:(NSUInteger)index;
```
mkchoi212: Replicates Swift STL's `func remove(at: Int) -> T `

Sources/VLCMediaList.m
```diff
@@ -217,11 +217,6 @@ - (void)insertObject:(VLCMedia *)object inMediaAtIndex:(NSUInteger)i
     [self insertMedia:object atIndex:i];
 }
 
-- (void)removeObjectFromMediaAtIndex:(NSUInteger)i
```
mkchoi212, line 219: Seems like it's not being used anywhere

Sources/VLCMediaList.m
fkuehne: This will lead to an out-of-bounds exception and early application termination by the runtime if the index is not in range. Please add sanity checks.

Headers/Public/VLCMediaList.h
fkuehne: I don't understand why you want to return a VLCMedia instance that is gone from the list.

Sources/VLCMediaList.m
fkuehne: Correct.

Sources/VLCMediaList.m
```diff
@@ -315,14 +311,17 @@ - (void)mediaListItemAdded:(NSArray *)arrayOfArgs
...
+    
     dispatch_sync(_serialMediaObjectsQueue, ^{
-        [_mediaObjects removeObjectAtIndex:[index intValue]];
+        if ([_mediaObjects containsObject:deleted]) {
+            [_mediaObjects removeObject:deleted];
+        }
     });
-    [self didChange:NSKeyValueChangeRemoval valuesAtIndexes:[NSIndexSet indexSetWithIndex:[index intValue]] forKey:@"media"];
-
+
```
fkuehne: Why do you remove support for key-value coding here?

Sources/VLCMediaList.m
```diff
@@ -163,17 +164,18 @@ - (void)insertMedia:(VLCMedia *)media atIndex: (NSUInteger)index
...
-- (void)removeMediaAtIndex:(NSUInteger)index
+- (VLCMedia *)removeMediaAtIndex:(NSUInteger)index
 {
+    VLCMedia *target = [_mediaObjects objectAtIndex:index];
+    
     dispatch_sync(_serialMediaObjectsQueue, ^{
-        if (index >= [_mediaObjects count])
-            return;
-        //remove from cached Media
+        // Remove from cached Media
```
fkuehne: hmm, why do you remove the sanity check here? This is not a fail-safe operation.

Sources/VLCMediaList.m
mkchoi212: I personally think this is the desirable behavior for the end user. I had a hard time debugging things when `removeMediaAtIndex` function wasn't critically failing even when an invalid index was passed to it.

Headers/Public/VLCMediaList.h
mkchoi212: Swift STL's `remove(at:)` behaves the same way and if the function doesn't returns `void`, how do I easily tell if a media had been deleted or not? A way to do is to raise NSRangeException, which leads to the use of `indexOf` without sanity checks in the function. Sorry if I'm not making sense 😅

Sources/VLCMediaList.m
mkchoi212: Ah I forgot to ask you about this. I personally didn't know you could do this until I saw the code for it. So, is this a feature that is used by users when they already have Notifications and delegates they can check?

Sources/VLCMediaList.m
fkuehne: No, the library shouldn't crash due to a potentially inconsistent state.

Headers/Public/VLCMediaList.h
fkuehne: Considering BOOL as a return value to expose the successful deletion. This matches typical AppKit behavior.

Sources/VLCMediaList.m
fkuehne: This is a used feature throughout multiple places of VLCKit, notably VLCMedia and VLCMediaPlayer. Especially the VLC-iOS app uses it.

Headers/Public/VLCMediaList.h
```diff
@@ -104,12 +104,13 @@ extern NSString *const VLCMediaListItemDeleted;
 - (void)insertMedia:(VLCMedia *)media atIndex:(NSUInteger)index;
 
 /**
- * remove a media from a given position
+ * remove and return the boolean result of the operation
```
carolanitz: remove media at position index and return true if the operation was successful. An unsuccessful operation occurs when the index is greater than the medialists count

Sources/VLCMedia.m
```diff
@@ -300,6 +300,12 @@ - (NSComparisonResult)compare:(VLCMedia *)media
     return p_md == [media libVLCMediaDescriptor] ? NSOrderedSame : NSOrderedAscending;
 }
 
+- (BOOL)isEqual:(id)other
+{
+    return ([other isKindOfClass: [VLCMedia class]] &&
+            [other libVLCMediaDescriptor] == p_md);
```
carolanitz, line 306: 👍

Sources/VLCMediaList.m
```diff
@@ -185,10 +191,9 @@ - (VLCMedia *)mediaAtIndex:(NSUInteger)index
     return media;
 }
 
-- (NSInteger)indexOfMedia:(VLCMedia *)media
+- (NSUInteger)indexOfMedia:(VLCMedia *)media
 {
-    NSInteger result = libvlc_media_list_index_of_item(p_mlist, [media libVLCMediaDescriptor]);
-    return result;
+    return [_mediaObjects indexOfObject:media];
```
carolanitz, line 196: Do all these changes to indexOfMedia and removeMediaAtIndex work with discovered media in a network as well? Being in a WLAN with a NAS or twonky server for example should be enough to test these changes

Sources/VLCMediaList.m
```diff
@@ -315,14 +315,19 @@ - (void)mediaListItemAdded:(NSArray *)arrayOfArgs
     [self didChange:NSKeyValueChangeInsertion valuesAtIndexes:[NSIndexSet indexSetWithIndexesInRange:range] forKey:@"media"];
 }
 
-- (void)mediaListItemRemoved:(NSNumber *)index
+- (void)mediaListItemRemoved:(NSArray *)arguments
 {
+    NSNumber *index = arguments[0][@"index"];
+    VLCMedia *deleted = arguments[0][@"media"];
```
carolanitz: can you use .first or firstobject instead of [0]?

Headers/Public/VLCMediaList.h
mkchoi212: Well said 👍

Sources/VLCMediaList.m
mkchoi212: I think so. There is no reason why it shouldn't. And not really sure how to test

> Being in a WLAN with a NAS or twonky server 

😅

carolanitz: you can try it for 30 days :) https://twonky.com/

carolanitz: and then you set it up build vlc ios with the custom vlckit and just check as it discovers files that all files  are discovered and that it doesn't crash somewhere unexpected and when you delete one out of the discovered folder that it all works as expected

mkchoi212: Just confirmed it still works and found [a bug](https://code.videolan.org/videolan/vlc-ios/issues/283) while at it :D

carolanitz (approved):
looks really good!

carolanitz:
this was merged with 005bcd708664a6671ad74a4f82ca447b1a8383df & 759bad2a90b3a5c77e89403f49a3a69e617d1572 & 0735d2b17c49099a9ffedc15e25eedf10a35a8c4 & 64130802fa4f4a7e1e5e5025005ac0dc4e665c36

PR #35, opened 2018-08-01
Tests: VLCMediaList

mkchoi212:
Includes tests for #34 `VLCMediaList::removeMediaAt`
Needs to be rebased after #34

### DIFF
--- a/Tests/Sources/VLCMediaListDelegateTest.swift
+++ b/Tests/Sources/VLCMediaListDelegateTest.swift
@@ -76,7 +76,8 @@ class VLCMediaListDelegateTest: XCTestCase {
             (10, 0, false),
         ]
         
-        let source = Video.standards.map{ VLCMedia(path: $0.path) }
+        let videos = [Video.test1, Video.test2, Video.test3, Video.test4]
+        let source = videos.map{ VLCMedia(path: $0.path) }
         let mediaList = try XCTAssertNotNilAndUnwrap(VLCMediaList(array: source))
 
         for (deleteIdx, count, removalSuccessful) in tests {
--- a/Tests/Sources/VLCMediaListDelegateTest.swift
+++ b/Tests/Sources/VLCMediaListDelegateTest.swift
@@ -1,0 +1,99 @@
+/*****************************************************************************
+ * VLCMediaListDelegateTest.swift
+ *****************************************************************************
+ * Copyright (C) 2018 Mike JS. Choi
+ * Copyright (C) 2018 VLC authors and VideoLAN
+ * $Id$
+ *
+ * Authors: Mike JS. Choi <mkchoi212 # icloud.com>
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation; either version 2.1 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston MA 02110-1301, USA.
+ *****************************************************************************/
+
+import XCTest
+
+final class MockMediaListDelegate: NSObject, VLCMediaListDelegate {
+    var removedIdx: UInt?
+    var removedExpectation: XCTestExpectation?
+    
+    func mediaList(_ aMediaList: VLCMediaList!, mediaRemovedAt index: UInt) {
+        removedIdx = index
+        removedExpectation?.fulfill()
+    }
+}
+
+class VLCMediaListDelegateTest: XCTestCase {
+    
+    // MARK: Delegate callbacks
+    
+    func testMediaRemovedAt() throws {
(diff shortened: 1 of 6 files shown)


Tests/Sources/VLCMediaListDelegateTest.swift
```diff
@@ -0,0 +1,97 @@
...
+        mediaList.delegate = delegate
+        
+        for idx in stride(from: UInt(source.count - 1), to: 0, by: -1) {
+            let delegateCalled = expectation(description: "delegate::mediaRemovedAt called")
+            delegate.removedExpectation = delegateCalled
+            
+            let oldCount = mediaList.count
+            
+            // Remove media from list
+            let ok = mediaList.removeMedia(at: idx)
```
carolanitz, line 56: move the XCTAssertTrue here maybe? :D

Tests/Sources/VLCMediaListDelegateTest.swift
```diff
@@ -0,0 +1,97 @@
...
+            (0,  0, true),
+            (10, 0, false),
+        ]
+        
+        let source = Video.standards.map{ VLCMedia(path: $0.path) }
+        let mediaList = try XCTAssertNotNilAndUnwrap(VLCMediaList(array: source))
+
+        for (deleteIdx, count, ok) in tests {
+            let removalResult = mediaList.removeMedia(at: deleteIdx)
+            XCTAssertEqual(removalResult, ok)
```
carolanitz: I would rename it from ok to removalSuccessful

Tests/Sources/VLCMediaListDelegateTest.swift
```diff
@@ -0,0 +1,97 @@
...
+            (9,  4, false),
+            (3,  3, true),
+            (1,  2, true),
+            (99, 2, false),
+            (1,  1, true),
+            (0,  0, true),
+            (10, 0, false),
+        ]
+        
+        let source = Video.standards.map{ VLCMedia(path: $0.path) }
```
carolanitz: isn't this fragile? When someone adds a Video to the standards the entire tests will fail, right?

Tests/Sources/VLCMediaListTest.swift
```diff
@@ -0,0 +1,55 @@
...
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston MA 02110-1301, USA.
+ *****************************************************************************/
+
+import XCTest
+
+class VLCMediaListTest: XCTestCase {
+
+    func testRemoveMedia() throws {
+        let videos = [Video.test1, Video.test1, Video.test2, Video.test3, Video.test4, Video.test4]
```
carolanitz, line 30: I like this approach better than the standards

Tests/Sources/VLCMediaListTest.swift
```diff
@@ -0,0 +1,55 @@
...
+            (2, [NF, 0, 1,NF,NF, 2], true),
+            (1, [NF, 0,NF,NF,NF, 1], true),
+            (8, [NF, 0,NF,NF,NF, 1], false),
+            (0, [NF,NF,NF,NF,NF, 0], true),
+            (0, [NF,NF,NF,NF,NF,NF], true),
+            (6, [NF,NF,NF,NF,NF,NF], false)
+        ]
+        
+        for (targetIdx, expectedState, ok) in tests {
+            let removalResult = mediaList.removeMedia(at: targetIdx)
```
carolanitz, line 48: 👍

Tests/Sources/VLCMediaListTest.swift
```diff
@@ -0,0 +1,55 @@
...
+        let NF = UInt(NSNotFound)
+        let tests: [(targetIdx: UInt, expectedState: [UInt], ok: Bool)] = [
+            (0, [NF, 0, 1, 2, 3, 4], true),
+            (3, [NF, 0, 1, 2,NF, 3], true),
+            (9, [NF, 0, 1, 2,NF, 3], false),
+            (2, [NF, 0, 1,NF,NF, 2], true),
+            (1, [NF, 0,NF,NF,NF, 1], true),
+            (8, [NF, 0,NF,NF,NF, 1], false),
+            (0, [NF,NF,NF,NF,NF, 0], true),
+            (0, [NF,NF,NF,NF,NF,NF], true),
```
carolanitz, line 43: quick question when you try again to remove at index 0 after removing it already at index 0 shouldn't the result be false? Do we need the NF in expectedState? I found it a bit confusing to match the index with the correct position because of it

mkchoi212:
Commits have been rebased :D Sorry for noticing it just now!

Tests/Sources/VLCMediaListTest.swift
mkchoi212: > quick question when you try again to remove at index 0 after removing it already at index 0 shouldn't the result be false?

No, because the previous state contains a media. `expectedState` is what the list looks like **after** the removal. 

> Do we need the NF in expectedState? 

I see where you are coming from and yeah it's a bit hard to match the indices. But I think it would be better to be explicit about exactly which media was removed.

e.g. Let say you have something like this

```swift
let videos = [Video.test1, Video.test1]
let removalResult = mediaList.removeMedia(at: 0)
let currentState = source.map{ mediaList.index(of: $0) }
XCTAssertEqual(currentState, [0])
```

From `XCTAssertEqual(currentState, [0])`, you can't tell which of the two `Video.test1` was removed.

MarcelRaschke (approved):

lawchingman (approved):

lawchingman (approved):

lawchingman (approved):

lawchingman (approved):

lawchingman (approved):

PR #36, opened 2018-08-02
Add carthage support by adding json files

carolanitz:


README.md
```diff
@@ -88,7 +88,25 @@ Then, run the following command,
 ```bash
 $ pod install
 ```
+### Carthage
 
+[Carthage](https://github.com/Carthage/Carthage) is  a way to add frameworks to your Cocoa application. You can install it with the following command,
```
mkchoi212: There are two spaces between `is` and `a way to` :D

Mikanbu: Additionally, I think we should add Carthage to the table of contents at the top.

fkuehne (approved):
Looks great! (assuming that Carthage supports tar.xz)

carolanitz:
since his was merged yes :D https://github.com/Carthage/Carthage/pull/2548

PR #38, opened 2018-08-06
VLCMedia: fix mediaWithURL redundant encoding (closes #139)

mkchoi212:
Gitlab [Issue](https://code.videolan.org/videolan/VLCKit/issues/139)

PR **checks if the provided url has been already encoded** to prevent redundant url encoding.

### DIFF
--- a/Sources/VLCMedia.m
+++ b/Sources/VLCMedia.m
@@ -269,13 +269,7 @@ static void HandleMediaParsedChanged(const libvlc_event_t * event, void * self)
         const char *url;
         VLCLibrary *library = [VLCLibrary sharedLibrary];
         NSAssert(library.instance, @"no library instance when creating media");
-
-        if (([[anURL absoluteString] hasPrefix:@"sftp://"]) ||
-            ([[anURL absoluteString] hasPrefix:@"smb://"])) {
-            url = [[[anURL absoluteString] stringByRemovingPercentEncoding] UTF8String];
-        } else {
-            url = [[anURL absoluteString] UTF8String];
-        }
+        url = [[[anURL absoluteString] stringByRemovingPercentEncoding] UTF8String];
 
         p_md = libvlc_media_new_location(library.instance, url);
 
--- a/Tests/Sources/VLCMediaTest.swift
+++ b/Tests/Sources/VLCMediaTest.swift
@@ -38,4 +38,41 @@ class VLCMediaTest: XCTestCase {
             XCTAssertEqual(expected, actual, input)
         }
     }
+    
+    func testInitWithUrl() throws {
+        let tests = [
+            "sftp://dummypath.mov",
+            "smb://dummypath.mkv",
+            "http://www.xyz.com/我们走吧.mp3".encodeURL(),
+            "smb://server/가즈아.mp3".encodeURL(),
+            "smb://server/media file.mp3".encodeURL()
+        ]
+        
+        for path in tests {
+            let url = try XCTAssertNotNilAndUnwrap(URL(string: path))
+            let media = VLCMedia(url: url)
+            XCTAssertEqual(media.url, url)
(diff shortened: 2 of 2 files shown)


Sources/VLCMedia.m
```diff
@@ -948,7 +948,9 @@ - (void)initInternalMediaDescriptor
         return;
     }
 
-    urlString = [urlString stringByAddingPercentEscapesUsingEncoding:NSUTF8StringEncoding];
+    if ([[urlString stringByRemovingPercentEncoding] isEqualToString:urlString]) {
+        urlString = [urlString stringByAddingPercentEscapesUsingEncoding:NSUTF8StringEncoding];
```
carolanitz: shouldn't this be stringByAddingPercentEncodingWithAllowedCharacters instead?

carolanitz (approved):
just that one deprecation change rest looks good :)

Sources/VLCMedia.m
```diff
@@ -948,7 +948,9 @@ - (void)initInternalMediaDescriptor
         return;
     }
 
-    urlString = [urlString stringByAddingPercentEscapesUsingEncoding:NSUTF8StringEncoding];
+    if ([[urlString stringByRemovingPercentEncoding] isEqualToString:urlString]) {
+        urlString = [urlString stringByAddingPercentEncodingWithAllowedCharacters:[NSCharacterSet URLQueryAllowedCharacterSet]];
```
mkchoi212: Not 100% sure about `URLQueryAllowedCharacterSet`. Thoughts?? @carolanitz

carolanitz: @tguillem just pointed out that we should never get into the situation where we deal with a percentencoded string in the first place and we should catch this case earlier when a user of the framework gives us such a url and adjust the documentation to make clear that the url should not be percentencoded

carolanitz (changes requested):
@tguillem just pointed out that we should never get into the situation where we deal with a percentencoded string in the first place and we should catch this case earlier when a user of the framework gives us such a url and adjust the documentation to make clear that the url should not be percentencoded

Sources/VLCMedia.m
```diff
@@ -269,13 +269,7 @@ - (instancetype)initWithURL:(NSURL *)anURL
...
         VLCLibrary *library = [VLCLibrary sharedLibrary];
         NSAssert(library.instance, @"no library instance when creating media");
-
-        if (([[anURL absoluteString] hasPrefix:@"sftp://"]) ||
-            ([[anURL absoluteString] hasPrefix:@"smb://"])) {
-            url = [[[anURL absoluteString] stringByRemovingPercentEncoding] UTF8String];
-        } else {
-            url = [[anURL absoluteString] UTF8String];
-        }
+        url = [[[anURL absoluteString] stringByRemovingPercentEncoding] UTF8String];
```
mkchoi212, line 272: @tguillem When creating a NSURL with special characters (e.g. Chinese characters), you need to percent-encode the string. This removes any existing percent encoding if the user passes in one to prevent duplicate percent-encoding in the future.

carolanitz:
@tguillem can one of you take another look at this, to check mikes change?

carolanitz:
needs testing before merging make sure you can still open http url with lots of special chars. alphanum + "-._~" should stay as is, everything else needs to be percent encoded. 
see for reference: http://git.videolan.org/?p=vlc.git;a=blob;f=src/text/url.c;h=3cfefb6aa1b2de5a5646e41bbfe77731142fb9dc;hb=HEAD#l86

carolanitz:
closing because we removed encoding handling entirely with 6b0b1794eb1b1ac802cbe815bf901b183a1078ad thank you very much though for this PR I tried it but had trouble opening some URLs. so we adjusted it to be the way the desktop apps do it

PR #40, opened 2018-08-18
Fixed issue of initializing VLCMedia with already opened stream

amosavian:
Pointer to `NSInputStream` was not passed to `open_cb`'s `opaque` parameter when stream is already opened by user.
Also removed reference to non-existing VLCHTTPInputStream

### DIFF
--- a/Sources/VLCMedia.m
+++ b/Sources/VLCMedia.m
@@ -70,14 +70,16 @@ NSString *const VLCMediaMetaChanged              = @"VLCMediaMetaChanged";
 int open_cb(void *opaque, void **datap, uint64_t *sizep) {
     NSInputStream *stream = (__bridge NSInputStream *)(opaque);
     
+    *datap = opaque;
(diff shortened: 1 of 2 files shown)


Sources/VLCMedia.m
```diff
@@ -76,8 +76,11 @@ int open_cb(void *opaque, void **datap, uint64_t *sizep) {
         *datap = opaque;
         *sizep = UINT64_MAX;
         return 0;
+    } else {
+        *datap = opaque;
```
mkchoi212: One minor suggestion! Lines 76, 77, 80, and 81 could be pulled outside the if-statement for more concise code :D

amosavian: Thanks I’ll do

carolanitz (approved):

carolanitz:
merged with 53695abfbd965c0f9be3af325cd4fbc7f3c56868

PR #41, opened 2018-08-18
Enabling iOS bitcode

amosavian:
This PR does two things:

- Enables bitcode for `DynamicMobileVLCKit` target.
- Makes `DynamicMobileVLCKit.h` public.

Bitcode and `-read_only_relocs` flag are incompatible when compiling framework for iOS device thus we have limited it to simulator only.


carolanitz (approved):
LGTM :+1: Thank you very much for the PR

PR #47, opened 2018-10-02
compileAndBuildVLCKit: Add three way merge for patches

Mikanbu:
This should ~sometimes~ make merging patches easier is there is any conflicts.


fkuehne:
I can't recommend going down this route as this just hides problems that will bite you later.

PR #49, opened 2018-10-03
VLCLibrary: Enable stats for iOS

Mikanbu:
Re-enable stats for iOS


fkuehne:
There is a performance penalty when using stats in libvlc, so this might only be useful in debug builds or should be optional.

PR #51, opened 2018-10-12
Font: fix defaultfontsize being zero and fall back to 10 in that case

carolanitz:
This fixes that subtitles are not visible by default since they have a negative size or zero


Mikanbu (approved):
LGTM.

PR #52, opened 2018-10-12
VLCMediaPlayer: VLCMediaPlayerStateToString: Return ES Added

Mikanbu:
Return`VLCMediaPlayerStateESAdded` also in order to be in sync with the enum.


fkuehne (approved):
Nicely spotted!

PR #53, opened 2018-10-15
VLCLibrary: Simplify HandleMessage

Mikanbu:
From man: vasprintf() will return -1 and set ret to be a NULL pointer.

### DIFF
--- a/Sources/VLCLibrary.m
+++ b/Sources/VLCLibrary.m
@@ -206,17 +206,9 @@ static void HandleMessage(void *data,
         return;
 
     char *str;
-    if (vasprintf(&str, fmt, args) == -1) {
-        if (str)
-            free(str);
-        str = NULL;
-        return;
+
+    if (vasprintf(&str, fmt, args) != -1) {
+        VKLog(@"%s", str);
+        free(str);
     }
-
-    if (str == NULL)
-        return;
-
-    VKLog(@"%s", str);
-    free(str);
-    str = NULL;
 }


fkuehne:
This will leak str if vasprintf fails but still allocates the pointer, no?

chouquette:
> This will leak str if vasprintf fails but still allocates the pointer, no?

It won't. In case of (non allocation related) failure asprintf is expected to release its buffer and to return -1. 
While the macOS wording is unclear, that's what is stated in the GNU/Linux doc: `If memory allocation wasn't possible, or some other error occurs, these functions will return -1, and the contents of strp are undefined.`

chouquette:
That being said, if the point is to simply pass the format string to VKLog, you might as well not allocate the buffer and simply do something like `VKLog(fmt, args)`

fkuehne:
There was something why I didn’t do it in such a simple way, but I’ll retry tonight.

Thanks a lot for the clarification on vasprintf and its memory handling!

chouquette:
Most likely the compiler complained about "fmt" not being a format string, which you could solve by flagging the function as a printf-like function as such: `static void [...] va_list args) __attribute__ ((format (printf, 4, 0)))`

fkuehne:
The default NSString format string operations can't work with va_list, so indeed the approach by @Mikanbu is absolutely correct and since it doesn't leak, it's a great simplification.

Optionally, I'd love to see an additional feature in this code path to reveal the code file and its line that send the message. This would add quite a bit of verbosity, but if provided through an option, it could be really helpful to debug things.

const char *module;
const char *file;
unsigned line;

libvlc_log_get_context(ctx, &module, &file, &line);

VKLog(@"%s/%s (%i): %s", module, file, line, str);
free(str);

... would do the trick.

Mikanbu:
Thanks for the reviews!

@fkuehne Indeed, I also wanted to add more info such as the module but I saw that module was always set to `libvlc`. 😢 

Additionally, @tguillem mentioned that ideally, we should have a module like android(`modules/logger/android.c`).

fkuehne:
Superseded by https://code.videolan.org/videolan/VLCKit/-/commit/beb9992608f25b29e82882676dc8e36f3a877768

PR #54, opened 2018-10-29
Feature recording

carolanitz:
This adds the functionality to start recording when a playing back a video.
It has callbacks for recording began and recording stopped
it also adds a function to then mux mp4 and srt into one file once recording is done.

This serves as a base for discussion.


Headers/Public/VLCMediaPlayer.h
```diff
@@ -875,6 +889,28 @@ extern NSString *const VLCTitleDescriptionIsMenu;
 @property (NS_NONATOMIC_IOSONLY, readonly) NSImage *lastSnapshot;
 #endif
 
+/**
+ * Start recording at given **directory** path
+ * \param path directory where the recording should go
+ * \return an BOOL with the success status
```
carolanitz, line 895: we should elaborate what failures can lead to a NO

Headers/Public/VLCMediaPlayer.h
```diff
@@ -875,6 +889,28 @@ extern NSString *const VLCTitleDescriptionIsMenu;
...
+ */
+- (BOOL)startRecordingAtPath:(NSString *)path;
+
+/**
+ * Stop current recording
+ * \return an BOOL with the success status
+ */
+- (BOOL)stopRecording;
```
carolanitz, line 903: when can recording fail?

Headers/Public/VLCMediaPlayer.h
```diff
@@ -875,6 +889,28 @@ extern NSString *const VLCTitleDescriptionIsMenu;
...
+
+/**
+ * mux srt and mp4 file to an mp4 file with embedded subtitles
+ * \param srtPath path to srt file
+ * \param mp4Path path to mp4 file
+ * \param outPath path where the new file should be written to
+ * \return an BOOL with the success status
+ */
+- (BOOL)muxSubtitleFile:(NSString *)srtPath toMp4File:(NSString *)mp4Path outputPath:(NSString *)outPath;
```
carolanitz: we should add nonnull to all params

Sources/VLCMediaPlayer.m
```diff
@@ -87,6 +87,8 @@ - (instancetype)initWithDrawable:(id)aDrawable options:(NSArray *)options;
 
 - (void)registerObservers;
 - (void)unregisterObservers;
+- (void)registerObserversForMuxWithPlayer:(libvlc_media_player_t *)player;
```
carolanitz: do these need to be extra functions?

Sources/VLCMediaPlayer.m
```diff
@@ -322,6 +351,9 @@ - (void)dealloc
     if (_viewpoint)
         libvlc_free(_viewpoint);
 
+    if (_p_mp)
+        libvlc_media_player_release(_p_mp);
```
carolanitz: this crashes sometimes

Sources/VLCMediaPlayer.m
```diff
@@ -1328,6 +1360,34 @@ - (void *)libVLCMediaPlayer
...
+}
+
+- (BOOL)stopRecording
+{
+    return libvlc_media_player_record(_playerInstance, NO, nil);
+}
+
+- (BOOL)muxSubtitleFile:(NSString *)srtPath toMp4File:(NSString *)mp4Path outputPath:(NSString *)outputPath
+{
+    //check if mp4 file is valid, srt is valid
```
carolanitz: I tried the following but somehow width and height directly didn't work
```suggestion
        if (![[NSFileManager defaultManager] fileExistsAtPath:srtPath] || ![[NSFileManager defaultManager] fileExistsAtPath:mp4Path]) {
        VKLog(@"no file at either srt or mp4 filepath");
        return NO;
    }
     libvlc_media_t* p_media = libvlc_media_new_path([_privateLibrary instance], [mp4Path UTF8String]);
    NSInteger width = 0;
    NSInteger height = 0;
    for (int i = 0; i < _media.tracksInformation.count; i++) {
        width = [_media.tracksInformation[i] integerForKey:VLCMediaTracksInformationVideoWidth];
        height = [_media.tracksInformation[i] integerForKey:VLCMediaTracksInformationVideoHeight];
    }

 

+    NSString *option = [NSString stringWithFormat:@":sout=#transcode{vcodec=h264,width=%ld,height=%ld,venc=avcodec{codec=h264_videotoolbox},acod(long)ec=mpga,ab=128,channels=2,samplerate=44100,soverlay}:file{dst='%@',mux=mp4}", width, height, outputPath];
```

carolanitz:
patches need to be rebased and have conflicts and this PR adds warnings which should be resolved before merging

Headers/Public/VLCMediaPlayer.h
fkuehne: well, and when can stopping recording fail?

Headers/Public/VLCMediaPlayer.h
```diff
@@ -875,6 +889,28 @@ extern NSString *const VLCTitleDescriptionIsMenu;
 @property (NS_NONATOMIC_IOSONLY, readonly) NSImage *lastSnapshot;
 #endif
 
+/**
+ * Start recording at given **directory** path
+ * \param path directory where the recording should go
```
fkuehne, line 894: Why is this a path and not a local file URL?

Headers/Public/VLCMediaPlayer.h
fkuehne: IMO, no. We should assume nonnull for the entire header. Everything else is just cluttering the code.

fkuehne: I fear that this method does not really belong in a playback class as it is completely separate from playing anything (even though the inside of libvlc looks different). Shouldn't this be part of a class designated to converting, muxing or streaming media?

Resources/MobileVLCKit/patches/0006-contrib-gcrypt-work-around-a-libtool-limitation.patch
```diff
@@ -31,27 +31,26 @@ index 0000000000..df97ffb488
...
 ++LIBTOOL=@LIBTOOL@ --tag=CC
 ++
 + AM_CPPFLAGS = -I../src -I$(top_srcdir)/src
 + AM_CFLAGS = $(GPG_ERROR_CFLAGS)
-+ 
--- 
++
+--
 2.19.1
-
```
fkuehne: Are you sure that deleting those white spaces does not break merging the patch? I remember having kept them for a reason.

Resources/MobileVLCKit/patches/0008-contrib-gcrypt-update-patches.patch
```diff
@@ -55,90 +60,90 @@ index f1d3ccc71f..21d7a30c8d 100644
...
++	movq RW0, (RW2);
++
 +-	call _gcry_3des_amd64_crypt_blk3;
 ++	call __gcry_3des_amd64_crypt_blk3;
-+ 
-+ 	popq %rdx; /*src*/
-+ 	popq %rsi; /*dst*/
++
++	popq %rdx; /*src*/
++	opq %rsi; /*dst*/
```
fkuehne: Are you sure this change is correct?

Resources/MobileVLCKit/patches/0008-contrib-gcrypt-update-patches.patch
```diff
@@ -360,8 +362,8 @@ index df97ffb488..957f268ad2 100644
 +--- libgcrypt/mpi/Makefile.am	2016-03-23 12:59:34.000000000 +0100
 ++++ libgcrypt/mpi/Makefile.am	2016-09-12 15:13:28.000000000 +0200
  @@ -23,6 +23,9 @@
-  
+ 
   # Need to include ../src in addition to top_srcdir because gcrypt.h is
 -- 
-2.19.1
+2.17.1
```
fkuehne: In conclusion, I don't understand why you update this patch nor why or what you are attempting to change. Pretty sure, removing the white spaces will break things.

Resources/MobileVLCKit/patches/0024-access_output-file-Add-error-dialog-for-write-open.patch
```diff
@@ -0,0 +1,40 @@
...
+index fbefce0be8..a2a72db411 100644
+--- a/modules/access_output/file.c
++++ b/modules/access_output/file.c
+@@ -88,6 +88,9 @@ static ssize_t Write( sout_access_out_t *p_access, block_t *p_buffer )
+         {
+             if (errno == EINTR)
+                 continue;
++            if (errno == ENOSPC)
++                vlc_dialog_display_error(p_access, "record",
++                                         "error: %s", vlc_strerror_c(errno));
```
fkuehne: Do we really want an end-user visible error here?

Sources/VLCMediaPlayer.m
fkuehne: Yes, they should be IMO.

Sources/VLCMediaPlayer.m
```diff
@@ -219,6 +247,7 @@ @interface VLCMediaPlayer ()
     BOOL _equalizerEnabled;                     ///< Equalizer state
     libvlc_video_viewpoint_t *_viewpoint;       ///< Current viewpoint of the media
     dispatch_queue_t _libVLCBackgroundQueue;    ///< Background dispatch queue to call libvlc
+    libvlc_media_player_t *_p_mp;
```
fkuehne: This is wrong. Use _playerInstance.

Sources/VLCMediaPlayer.m
```diff
@@ -1328,6 +1360,34 @@ - (void *)libVLCMediaPlayer
...
+    //check if mp4 file is valid, srt is valid
+
+    libvlc_media_t* p_media = libvlc_media_new_path([_privateLibrary instance], [mp4Path UTF8String]);
+
+    NSString *option = [NSString stringWithFormat:@":sout=#transcode{vcodec=h264,width=720,height=480,venc=avcodec{codec=h264_videotoolbox},acodec=mpga,ab=128,channels=2,samplerate=44100,soverlay}:file{dst='%@',mux=mp4}", outputPath];
+
+    libvlc_media_add_option(p_media, [[NSString stringWithFormat:@":sub-file=%@", srtPath] UTF8String]);
+    libvlc_media_add_option(p_media, [option UTF8String]);
+
+    _p_mp = libvlc_media_player_new_from_media( p_media );
```
fkuehne: Use _playerInstance.

Sources/VLCMediaPlayer.m
fkuehne: Are you use that tracksInformation is filled with content at this point? Additionally, this will lead to issues if the MP4 file being used to receive the subtitles does not match the media that is currently referenced, which will be the case quite often.
Further, your loop will fail if there is more than 1 video track in the given media (namely any further audio, video or SPU track).

Sources/VLCMediaPlayer.m
```diff
@@ -1328,6 +1360,34 @@ - (void *)libVLCMediaPlayer
...
+- (BOOL)stopRecording
+{
+    return libvlc_media_player_record(_playerInstance, NO, nil);
+}
+
+- (BOOL)muxSubtitleFile:(NSString *)srtPath toMp4File:(NSString *)mp4Path outputPath:(NSString *)outputPath
+{
+    //check if mp4 file is valid, srt is valid
+
+    libvlc_media_t* p_media = libvlc_media_new_path([_privateLibrary instance], [mp4Path UTF8String]);
```
fkuehne: Missing sanity check. What if mp4Path is nil?

Sources/VLCMediaPlayer.m
```diff
@@ -1328,6 +1360,34 @@ - (void *)libVLCMediaPlayer
...
+
+    NSString *option = [NSString stringWithFormat:@":sout=#transcode{vcodec=h264,width=720,height=480,venc=avcodec{codec=h264_videotoolbox},acodec=mpga,ab=128,channels=2,samplerate=44100,soverlay}:file{dst='%@',mux=mp4}", outputPath];
+
+    libvlc_media_add_option(p_media, [[NSString stringWithFormat:@":sub-file=%@", srtPath] UTF8String]);
```
fkuehne: Missing sanity check. You should fail early is srtPath is not provided or invalid.

Sources/VLCMediaPlayer.m
```diff
@@ -1328,6 +1360,34 @@ - (void *)libVLCMediaPlayer
...
+    return libvlc_media_player_record(_playerInstance, NO, nil);
+}
+
+- (BOOL)muxSubtitleFile:(NSString *)srtPath toMp4File:(NSString *)mp4Path outputPath:(NSString *)outputPath
+{
+    //check if mp4 file is valid, srt is valid
+
+    libvlc_media_t* p_media = libvlc_media_new_path([_privateLibrary instance], [mp4Path UTF8String]);
+
+    NSString *option = [NSString stringWithFormat:@":sout=#transcode{vcodec=h264,width=720,height=480,venc=avcodec{codec=h264_videotoolbox},acodec=mpga,ab=128,channels=2,samplerate=44100,soverlay}:file{dst='%@',mux=mp4}", outputPath];
```
fkuehne: Do we want to check whether outputPath exists or fail silently?
Additionally, consider renaming the variable as "option" is a bit generic.

Resources/MobileVLCKit/patches/0005-libvlc_media-add-cookie_jar-API.patch
```diff
@@ -88,7 +88,7 @@ index 54875d6a0e..e1d23caaf5 100644
  
 +    if( p_md->p_cookie_jar )
 +        vlc_http_cookies_destroy( p_md->p_cookie_jar );
-+    
++
```
fkuehne: See my other remarks on spaces.

Resources/MobileVLCKit/patches/0008-contrib-gcrypt-update-patches.patch
```diff
@@ -1,24 +1,29 @@
-From 85cd4dcac702493f4eb4c7e09dbf6914e2bfdb51 Mon Sep 17 00:00:00 2001
-From: =?UTF-8?q?Felix=20Paul=20K=C3=BChne?= <fkuehne@videolan.org>
-Date: Mon, 12 Sep 2016 17:03:37 +0200
-Subject: [PATCH 08/21] contrib/gcrypt: update patches
+From f1b6287eca313c6b438b08af20effd4c7e63a385 Mon Sep 17 00:00:00 2001
+From: Carola Nitz <nitz.carola@googlemail.com>
+Date: Tue, 30 Oct 2018 16:40:32 +0100
+Subject: [PATCH 08/24] contrib/gcrypt: update patches
```
fkuehne, line 4: Why are you stripping my copyright?

Resources/MobileVLCKit/patches/0024-access_output-file-Add-error-dialog-for-write-open.patch
```diff
@@ -0,0 +1,40 @@
...
+@@ -305,8 +308,13 @@ static int Open( vlc_object_t *p_this )
+             if (fd != -1)
+                 break;
+             if (fd == -1)
++            {
+                 msg_Err (p_access, "cannot create %s: %s", path,
+                          vlc_strerror_c(errno));
++
++                vlc_dialog_display_error(p_access, "record",
++                                         "error: %s", vlc_strerror_c(errno));
```
fkuehne: Idem.

Headers/Public/VLCMediaPlayer.h
carolanitz: The file name is generated from the url and a timestamp. Let's discuss what design might be better
If we let the user give us a file name we have to check if a file might already exists and handle that as well. Right now this is unlikely since we generate the recording from the url and add a timestamp. once recording is done an event tells you where the recording landed and you can move it or rename it afterwards

Headers/Public/VLCMediaPlayer.h
carolanitz: that is a good point!

carolanitz: I can add Nonnull for everything? are you entirely sure that some classes don't work with nil?
I didn't want to open that rabbit hole as part of this patch but yes good idea in general

Resources/MobileVLCKit/patches/0005-libvlc_media-add-cookie_jar-API.patch
carolanitz: they are not needed and the patch applies. This also resolves the "patch adds trailing whitespaces" warnings

Resources/MobileVLCKit/patches/0006-contrib-gcrypt-work-around-a-libtool-limitation.patch
carolanitz: yes very sure. This is not a whitespace sensitive language

Resources/MobileVLCKit/patches/0008-contrib-gcrypt-update-patches.patch
carolanitz: I regenerated the patch and that was done by accident

Resources/MobileVLCKit/patches/0008-contrib-gcrypt-update-patches.patch
carolanitz: no this is not correct. Good catch!

Resources/MobileVLCKit/patches/0008-contrib-gcrypt-update-patches.patch
carolanitz: after rebasing this patch didn't apply anymore. The whitespaces don't break anything but the missing p from popq certainly

Resources/MobileVLCKit/patches/0024-access_output-file-Add-error-dialog-for-write-open.patch
carolanitz: According to Thomas there is no way to propagate Errors, which would have been the ideal solution here. This is an error that says that the user doesn't have enough space to do their recording. Since this is not something you can check in advance but is easy to fix on the user side we decided to make that one user facing

Sources/VLCMediaPlayer.m
carolanitz: yes we already removed it because if you remove width and height it uses the default values of the stream

Resources/MobileVLCKit/patches/0008-contrib-gcrypt-update-patches.patch
carolanitz: there must be something off in my repository though because 3.0 compiles fine. I will double check

Headers/Public/VLCMediaPlayer.h
fkuehne: I understand that rationale and that makes sense. My initial question was though why this is path and not a URL to match AppKit/Foundation in their more recent API designs.

Resources/MobileVLCKit/patches/0006-contrib-gcrypt-work-around-a-libtool-limitation.patch
fkuehne: You misunderstood me. The point I wanted to bring across is about merging, not about the actual content.
Until very recently (and it can still be the case with current public releases), the "patch" utility on macOS is often failing if spaces in the original file are not reflected in the diff. This problem does not occur on Linux, but was biting quite often on the Mac. Hence, while git still complains, it is probably best to keep the spaces. Please try this thoroughly before you merge.

Resources/MobileVLCKit/patches/0024-access_output-file-Add-error-dialog-for-write-open.patch
fkuehne: Ok, however, in reality, the user will get a dialog stating "error: -13" without further explanation, but this is a limitation of libvlc, not of your patch.

Sources/VLCMediaPlayer.m
carolanitz: You are aware that the _playerinstance is playing back a stream, so you can not use the same instance for transcoding later on? Or am I missing something

fkuehne: I believe it would violate the state machine if you handle two player instances within a single instance of VLCMediaPlayer doing different things at the same time or consecutively. Especially event handling will be in the worst case undefined or at least hard to follow in the best case.
Following my earlier remark, I don't believe that transcoding should be part of the VLCMediaPlayer class at all, but maybe in a designated subclass.

Resources/MobileVLCKit/patches/0008-contrib-gcrypt-update-patches.patch
fkuehne: Let me know if you need help.

Headers/Public/VLCMediaPlayer.h
fkuehne: We should move that to different PR indeed. I believe that legitimate use of nil is pretty rare throughout VLCKit indeed.

Headers/Public/VLCMediaPlayer.h
carolanitz: Right now it returns no when there is no input stream. That doesn't seem right because you could be not recording and stopRecording would return NO as in it was not able to stop recording. Someone  from the core needs to look over the patches. I'm not sure if this is correctly implemented on the core side

Headers/Public/VLCMediaPlayer.h
carolanitz: okay just to clarify, you don't care about the local file part but about it being an NSURL instead of an NSString, correct? in that case I'm happy to change it to startRecordingAtDirectoryURL:(NSURL *)directoryURL

Resources/MobileVLCKit/patches/0024-access_output-file-Add-error-dialog-for-write-open.patch
carolanitz: I wasn't aware that it's "error: -13" that's actually not okay in that case and need to be addressed

Sources/VLCTranscoder.m
```diff
@@ -0,0 +1,120 @@
...
+        [self unregisterObserversForMuxWithPlayer:_p_mp];
+        libvlc_media_player_stop( _p_mp );
+    }
+}
+
+static void HandleMuxMediaInstanceStateChanged(const libvlc_event_t * event, void * self)
+{
+    VLCMediaPlayerState newState;
+
+    if (event->type == libvlc_MediaPlayerPaused) {
```
alexandre-janniaux: I believe you want to listen to the recording player and not the transcoding player, so that it does pause when the record pause right ? 
You might need a `pause()` method instead, as the event will be provided outside of this class.

But you can keep this, so you can listen to transcoding error event.

carolanitz: I removed that part from this PR

Headers/Public/VLCMediaPlayer.h
```diff
@@ -135,6 +136,19 @@ extern NSString * VLCMediaPlayerStateToString(VLCMediaPlayerState state);
  */
 - (void)mediaPlayerSnapshot:(NSNotification *)aNotification;
 
+/**
+ * Sent by the default notification center whenever the player started recording.
+ * @param player the player who started recording
+ */
+- (void)mediaPlayerRecordStarted:(VLCMediaPlayer *)player;
```
fkuehne, line 143: English grammar nag nag. Shouldn't this be "mediaPlayerStartedRecording"?

fkuehne (approved):
One nit-pick, rest of series looks top!

Headers/Public/VLCMediaPlayer.h
carolanitz: yes that would indeed sound nicer

tguillem:
PR fine with me, but we need to clarify record error handling for VLC 5.0.

carolanitz:
merged with 4bcb1a6d09b6a0d7a8aa910ee7065dd4678ecd91

PR #56, opened 2018-11-22
MediaPlayer: expose callback, rename callback

carolanitz:
mediaPlayerRecordChanged was not exposed which resulted in a warning
and recordStoppedAtPath -> recordingStoppedAtPath for consistency

### DIFF
--- a/Headers/Public/VLCMediaPlayer.h
+++ b/Headers/Public/VLCMediaPlayer.h
@@ -147,7 +147,7 @@ extern NSString * VLCMediaPlayerStateToString(VLCMediaPlayerState state);
  * @param player the player who stopped recording
  * @param path the path to the file that the player recorded to
  */
-- (void)mediaPlayer:(VLCMediaPlayer *)player recordStoppedAtPath:(NSString *)path;
+- (void)mediaPlayer:(VLCMediaPlayer *)player recordingStoppedAtPath:(NSString *)path;
 
 @end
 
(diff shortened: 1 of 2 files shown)


Headers/Public/VLCMediaPlayer.h
```diff
@@ -147,7 +147,13 @@ extern NSString * VLCMediaPlayerStateToString(VLCMediaPlayerState state);
...
  * @param path the path to the file that the player recorded to
  */
-- (void)mediaPlayer:(VLCMediaPlayer *)player recordStoppedAtPath:(NSString *)path;
+- (void)mediaPlayer:(VLCMediaPlayer *)player recordingStoppedAtPath:(NSString *)path;
+
+/**
+ * Sent by the default notification center whenever the player recording changed.
+ * @param arguments first ojects contains the "filePath" and the "isRecording" field
+ */
+- (void)mediaPlayerRecordChanged:(NSArray *)arguments;
```
carolanitz: I just noticed the arguments here and have to say that I'd rather have the bool and filepath separatly.
@Mikanbu  Is there any reason it's arguments in the first place ?

Mikanbu: IIRC, it's based on the `VLCEventManager` class and this method shouldn't be exposed.

carolanitz: also why is this an array that contains a dict and not immediatly a dict?

carolanitz: @Mikanbu ah I found the correct place then

Mikanbu (approved):
LGTM, I would just separate the changes into two commits!

carolanitz:
closed with 54b68f2d77f92369835268abba5fc608061822a1 & 5628ef2a3f8070b42339655439f28c98dae67b6d

PR #58, opened 2018-11-28
Bump MobileVLCKit and TVVLCKIT to 8.0 and 10.2 minimum deployment target

Mikanbu:
This fixes the MobileVLCKit nightly.


fkuehne (changes requested):
Please bump to iOS 8.4 since this is what we actually require. For tvOS, 10.2 is good!

Mikanbu:
Thank you for the review, I will make the change! Just out of curiosity should I bump to iOS 8.4 on 3.0 as well?

fkuehne:
Yes, absolutely. We require 8.4 since the introduction of ChromeCast support (this is already reflected in the podspec).

Mikanbu:
Oh I see, I will update the PR!

fkuehne:
Feel free to directly push with those changes!

PR #59, opened 2018-12-03
Add transcoding api

carolanitz:
This adds the Transcoding class and a function to mux subtitle files into a video

### DIFF
--- a/Headers/Public/VLCTranscoder.h
+++ b/Headers/Public/VLCTranscoder.h
@@ -1,0 +1,67 @@
+/*****************************************************************************
+ * VLCTranscoder.h: VLCKit.framework VLCTranscoder implementation
+ *****************************************************************************
+ * Copyright (C) 2018 Carola Nitz
+ * Copyright (C) 2018 VLC authors and VideoLAN
+ * $Id$
+ *
+ * Authors:  Carola Nitz <caro # videolan.org>
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation; either version 2.1 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston MA 02110-1301, USA.
+ *****************************************************************************/
+
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+* Transcoder delegate allows to be notified about transcoding state
+*/
+@class VLCTranscoder;
+@protocol VLCTranscoderDelegate <NSObject>
+
+@optional
+/**
+ * Called when the transcoding finished
+ * \param transcoder the transcoder object that finished
+ * \param success if transcoding finished sucessfully or not
+ */
+- (void)transcode:(VLCTranscoder *)transcoder finishedSucessfully:(BOOL)success;
+
+@end
+
+/**
+ * Provides an object to convert a subtitle file and moviefile into one.
+ */
+@interface VLCTranscoder: NSObject
+
+/**
+ * the delegate object implementing the optional protocol
+ */
+@property (weak, nonatomic) id<VLCTranscoderDelegate> delegate;
+
+/**
+ * Reencode and remuxes an srt and mp4 file to an mkv file with embedded subtitles either with VideoToolbox-based H264 encoding or VP80 is Videotoolbox is not available
+ * \param srtPath path to srt file
+ * \param mp4Path path to mp4 file
+ * \param outPath path where the new file should be written to
+ * \return an BOOL with the success status, returns NO if the subtitle file is not an srt or mp4File is not an mp4 file or the files don't exist at that path or transcoding failed for other reasons
+ */
+
+- (BOOL)reencodeAndMuxSRTFile:(NSString *)srtPath toMP4File:(NSString *)mp4Path outputPath:(NSString *)outPath;
+
+@end
+
+NS_ASSUME_NONNULL_END
--- a/Sources/VLCTranscoder.m
+++ b/Sources/VLCTranscoder.m
@@ -1,0 +1,143 @@
+/*****************************************************************************
+ * VLCTranscoder.m: VLCKit.framework VLCTranscoder implementation
+ *****************************************************************************
+ * Copyright (C) 2018 Carola Nitz
+ * Copyright (C) 2018 VLC authors and VideoLAN
+ * $Id$
+ *
+ * Authors:  Carola Nitz <caro # videolan.org>
+ *
+ * This program is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation; either version 2.1 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston MA 02110-1301, USA.
+ *****************************************************************************/
+#import "VLCTranscoder.h"
+
+#import "VLCEventManager.h"
+#import "VLCLibrary.h"
+#import "VLCLibVLCBridging.h"
+
+#include <vlc/vlc.h>
+
+@interface VLCTranscoder()
+{
+    libvlc_media_player_t *_p_mp; //player instance used for transcoding
+    dispatch_queue_t _libVLCTranscoderQueue;
+}
+- (void)registerObserversForMuxWithPlayer:(libvlc_media_player_t *)player;
+- (void)unregisterObserversForMuxWithPlayer:(libvlc_media_player_t *)player;
+@end
+
+@implementation VLCTranscoder
+
+- (instancetype)init
+{
+    if (self = [super init]) {
+        _libVLCTranscoderQueue = dispatch_queue_create("libVLCTranscoderQueue", DISPATCH_QUEUE_SERIAL);
+    }
+    return self;
+}
+
+- (BOOL)reencodeAndMuxSRTFile:(NSString *)srtPath toMP4File:(NSString *)mp4Path outputPath:(NSString *)outPath
+{
+    libvlc_media_t* p_media = libvlc_media_new_path([[VLCLibrary sharedLibrary] instance], [mp4Path UTF8String]);
+    if (p_media == NULL) {
+        NSAssert(0, @"p_media wasn't allocated");
+        return NO;
+    }
+    NSString *transcodingOptions = [NSString stringWithFormat:@":sout=#transcode{venc={module=avcodec{codec=h264_videotoolbox}, vcodec=h264},venc={module=vpx{quality-mode=2},vcodec=VP80},samplerate=44100,soverlay}:file{dst='%@',mux=mkv}", outPath];
+    libvlc_media_add_option(p_media, [[NSString stringWithFormat:@"--sub-file=%@", srtPath] UTF8String]);
+    libvlc_media_add_option(p_media, [transcodingOptions UTF8String]);
+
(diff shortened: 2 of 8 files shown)


Sources/VLCTranscoder.m
```diff
@@ -0,0 +1,142 @@
...
+{
+    if (self = [super init]) {
+        _libVLCTranscoderQueue = dispatch_queue_create("libVLCTranscoderQueue", DISPATCH_QUEUE_SERIAL);
+    }
+    return self;
+}
+
+- (BOOL)muxSubtitleFile:(NSString *)srtPath toMp4File:(NSString *)mp4Path outputPath:(NSString *)outputPath
+{
+    libvlc_media_t* p_media = libvlc_media_new_path([[VLCLibrary sharedLibrary] instance], [mp4Path UTF8String]);
```
carolanitz, line 53: srtPath and mp4Path check was removed since it is advised against

Headers/Public/VLCTranscoder.h
```diff
@@ -0,0 +1,65 @@
...
+* Transcoder delegate allows to be notified about transcoding state
+*/
+@protocol VLCTranscoderDelegate <NSObject>
+
+@optional
+/**
+ * Sent by the default notification center whenever the transcoder state has changed.
+ * \param aNotification
+ */
+- (void)transcoderFinishedSucessfully:(BOOL)success;
```
ePirat: The method name should probably be `transcodeFinishedSucessfully:`. The description of the method and documentation description are contradicting itself. The description implies it is called whenever the state has changed, the method name strongly implies it is called when finished successfully but then has a success argument, so maybe it is not finished successfully? Thats very confusing.

Headers/Public/VLCTranscoder.h
```diff
@@ -0,0 +1,65 @@
...
+
+@end
+
+/**
+ * Provides an object to convert a subtitle file and moviefile into one.
+ */
+@interface VLCTranscoder: NSObject
```
ePirat, line 48: Is this really a transcoder or a transcode? It seems to make sense to have a Transcoder to do transcode jobs but then it seems weird to have a delegate for the transcoder state and not the individual transcode jobs, how would you tell apart changes when you have multiple transcodes?

Headers/Public/VLCTranscoder.h
```diff
@@ -0,0 +1,65 @@
...
+
+/**
+ * mux srt and mp4 file to an mkv file with embedded subtitles either with VideoToolbox-based H264 encoding or VP80 is Videotoolbox is not available
+ * \param srtPath path to srt file
+ * \param mp4Path path to mp4 file
+ * \param outPath path where the new file should be written to
+ * \return an BOOL with the success status, returns NO if the subtitle file is not an srt or mp4File is not an mp4 file or the files don't exist at that path or transcoding failed for other reasons
+ */
+
+- (BOOL)muxSubtitleFile:(NSString *)srtPath toMp4File:(NSString *)mp4Path outputPath:(NSString *)outPath;
```
ePirat: Method name implies muxing happens but in reality it re-encodes and remuxes, as far as I understand the code, which is a very important difference so should be clarified. (Muxing is relatively fast, encoding not)

Nitpick: `toMP4File` instead of `toMp4File` and probably `muxSRTFile` if it only works for SRT.

Sources/VLCTranscoder.m
```diff
@@ -0,0 +1,142 @@
...
+    if (p_media == NULL) {
+        NSAssert(0, @"p_media wasn't allocated");
+        return NO;
+    }
+    NSString *transcodingOptions = [NSString stringWithFormat:@":sout=#transcode{venc={module=avcodec{codec=h264_videotoolbox}, vcodec=h264},venc={module=vpx{quality-mode=2},vcodec=VP80},acodec=mpga,ab=128,channels=2,samplerate=44100,soverlay}:file{dst='%@',mux=mkv}", outputPath];
```
ePirat: Not sure if hardcoding the audio codec to MPGA with only 128 kbits is a good idea? Doesn't that sound awful?

Sources/VLCTranscoder.m
```diff
@@ -0,0 +1,142 @@
...
+    libvlc_media_add_option(p_media, [transcodingOptions UTF8String]);
+
+    _p_mp = libvlc_media_player_new_from_media( p_media );
+    if (_p_mp == NULL) {
+        NSAssert(0, @"_p_mp wasn't allocated");
+        return NO;
+    }
+
+    [self registerObserversForMuxWithPlayer:_p_mp];
+    BOOL canPlay = libvlc_media_player_play( _p_mp ) == 0;
```
ePirat, line 69: `libvlc_media_player_play` returns int not BOOL, on macOS BOOL is a typedef for a signed char so this will probably produce a warning about implicitly casting to a smaller type, as usually char is smaller than int)

Sources/VLCTranscoder.m
```diff
@@ -0,0 +1,142 @@
...
+    }
+    return canPlay;
+}
+
+- (void)registerObserversForMuxWithPlayer:(libvlc_media_player_t *)player
+{
+    __block libvlc_event_manager_t * p_em = libvlc_media_player_event_manager(player);
+    if (!p_em)
+        return;
+    dispatch_sync(_libVLCTranscoderQueue,^{
```
ePirat, line 84: Why dispatch to the libVLCTranscoderQueue here but not when unregistering? Shouldn't just for safety self be retained here and released again only when unregistering to prevent that self is deallocated before having released the observers?

Sources/VLCTranscoder.m
```diff
@@ -0,0 +1,142 @@
...
+                        HandleMuxMediaInstanceStateChanged, (__bridge void *)(self));
+}
+
+- (void)mediaPlayerStateChangeForMux:(NSNumber *)newState
+{
+    if (_p_mp) {
+        [self unregisterObserversForMuxWithPlayer:_p_mp];
+        libvlc_media_player_stop( _p_mp );
+        if (self.delegate && [self.delegate respondsToSelector:@selector(transcoderFinishedSucessfully:)]) {
+            BOOL success = [newState isEqualToNumber: @(VLCMediaPlayerStateError)];
```
ePirat: `success` is the opposite of the intended meaning here, even though one line down it is negated this is a bit confusing when reading the code, better immediately negate the result of the `isEqualToNumber` method here when assigning and remove the negation below.

Sources/VLCTranscoder.m
```diff
@@ -0,0 +1,142 @@
...
+            [self.delegate transcoderFinishedSucessfully:!success];
+        }
+    }
+}
+
+static void HandleMuxMediaInstanceStateChanged(const libvlc_event_t * event, void * self)
+{
+    VLCMediaPlayerState newState;
+    if (event->type == libvlc_MediaPlayerPaused || event->type == libvlc_MediaPlayerEncounteredError || event->type == libvlc_MediaPlayerStopped) {
```
ePirat: I am having trouble understanding what the purpose of this check here is.

ePirat (commented):
Hi, thanks for your work! I guess the intention for this is currently mostly for transcoding for the Chromecast? I've left some comments inline.

Headers/Public/VLCTranscoder.h
carolanitz: oh I need to return the instance here of the transcoder object stupid me!

Headers/Public/VLCTranscoder.h
carolanitz: all good points!

Sources/VLCTranscoder.m
carolanitz: tbh this part of the code is not my area of expertise so can I just drop it?

ePirat: Ideally you would leave the audio alone here, that said depending on the source the audio codecs could maybe be incompatible with the MKV container.

carolanitz: so from speaking to @alexandre-janniaux I can't drop the audio options here

Sources/VLCTranscoder.m
carolanitz: We're checking that int against 0, if it's == 0 so it should be fine

Sources/VLCTranscoder.m
ePirat: So VLC has no way, like FFMpeg has, to "copy" a specific track to the output without transcoding it?

carolanitz: @alexandre-janniaux can you answer @ePirat because this is out of my knowledge scope

Sources/VLCTranscoder.m
carolanitz: Ohhhh we might have the same issue on the player :D

Sources/VLCTranscoder.m
carolanitz: very good point

Sources/VLCTranscoder.m
carolanitz: yeah it's obsolete looking at it

Sources/VLCTranscoder.m
alexandre-janniaux: There's no way for an elementary stream to avoid the transcode module but removing the audio parameters should tell the transcode module to avoid transcoding them and use them as-is.

I don't know the possible timestamp issue that could be raised by that, but it should be ok.

Sources/VLCTranscoder.m
ePirat: Indeed, completely overlooked that…

Headers/Public/VLCTranscoder.h
alexandre-janniaux: `blendSubtitleFile` or `blendSubtitleFileAndSave` might be a better name, as the subtitle is not muxed but burned into the video.

It should work with any subtitle file for which the decoder is available

Sources/VLCTranscoder.m
carolanitz: how would you retain self in this case?

Sources/VLCTranscoder.m
alexandre-janniaux: IIRC, it can be {EncounteredError, Stopped} on macosx, but it becomes {EncounteredError, Paused} on iOS. As the transcoder might never pause, this should be ok but we have to fix libvlc if so.

Headers/Public/VLCTranscoder.h
carolanitz: @alexandre-janniaux the problem is I don't know exactly which subtitle files are supported by the decoder. Because I would like to state that explicitly

Sources/VLCTranscoder.m
carolanitz: but still the check is obsolete because these are the only things I'm registering for so it will only be one of those three states

Sources/VLCTranscoder.m
carolanitz: I can try leaving it out

Sources/VLCTranscoder.m
ePirat:
```
void *retainedSelf = (__bridge_retained void *)self
```
and back

```
void *releaseSelf = (__bridge_transfer void *)self;
```
or `CFBridgingRetain` and `CFBridgingRelease`.

That said, if the lifetime is clear and this can not happen anyway, thats not needed.

Headers/Public/VLCTranscoder.h
alexandre-janniaux: it depends on which modules you build in libvlc when building vlckit. 

I'm not an expert in subtitles but I think you can get a rough idea of what subtitles are supported in vlc with `vlc -l | grep sub`, then you can expect that only LGPLv2 libraries are available on iOS, which would be everything but aribsub. I should have something like

```
  subtitle               Text subtitle parser
  demux_stl              EBU STL subtitles parser
  vobsub                 Vobsub subtitles parser
  stl                    EBU STL subtitles decoder
  dvbsub                 DVB subtitles decoder
  cvdsub                 CVD subtitle decoder
  spudec                 DVD subtitles decoder
  subsdec                Text subtitle decoder
  textst                 HDMV TextST subtitles decoder
  ttml                   TTML subtitles decoder
  svcdsub                Philips OGT (SVCD subtitle) decoder
  webvtt                 WEBVTT subtitles decoder
  subsusf                USF subtitles decoder
  substx3g               tx3g subtitles decoder
```

The list on videolan.org (vlc->features) is different but I don't know if it's up to date too. I might have missed some formats

```
 DVD, Text files (MicroDVD, SubRIP, SubViewer, SSA1-5, SAMI, VPlayer), Closed captions, Vobsub, Universal Subtitle Format (USF), SVCD / CVD, DVB, OGM, CMML, Kate. 
```

carolanitz: okay I'd say this is something we can add in an update later. Lets go for now with srt and get this shipped!

Sources/VLCTranscoder.m
carolanitz: I actually think you can get in the state where the object gets destroyed while transcoding
but that can be avoided by calling mediaPlayerStateChangeForMux on deallocation. That should be the only case where we have dangling event things. but on second thought what we want is the retain

carolanitz: but that can be avoided by calling mediaPlayerStateChangeForMux on deallocation

ePirat (approved):

PR #55, opened 2018-11-22
compileAndBuild: enable libmux

carolanitz:
closes #224

PR #60, opened 2018-12-06
compileAndBuildVLCKit: Remove libmux from blacklist

Mikanbu:


carolanitz (approved):
follow up to 525409e and https://github.com/videolan/vlckit/pull/55 👍

PR #61, opened 2018-12-06
VLCMediaList: Set QOS of _serialMediaObjectsQueue to background

Mikanbu:
In order to not block the UI for example in VLCMediaList `- (void)mediaListItemAdded:(NSArray *)arrayOfArgs`  l296, set the queue to background priority.

### DIFF
--- a/Sources/VLCMediaList.m
+++ b/Sources/VLCMediaList.m
@@ -93,7 +93,12 @@ static void HandleMediaListItemDeleted( const libvlc_event_t * event, void * use
 
         // Initialize internals to defaults
         _mediaObjects = [[NSMutableArray alloc] init];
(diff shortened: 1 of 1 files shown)


Mikanbu:
Any objections for this patch?

fkuehne:
If this does not have a noticeable impact on performance, LGTM.

carolanitz:
@fkuehne yeah I'm actually unsure since it sets the priority in which these tasks are done to a very low priority. Which means everthing else is way faster but these tasks might take a long time to be done @Mikanbu why did you decide for background instead of DISPATCH_QUEUE_PRIORITY_DEFAULT or DISPATCH_QUEUE_PRIORITY_LOW or even moving it off the mainthread ?

carolanitz:
@Mikanbu I believe it's QOS_CLASS_USER_INITIATED since it is related to events from the user like moving files onto the device or deleting a file. if that is changed I'm also good with it. see https://developer.apple.com/videos/play/wwdc2014/716/ minute 7 and following

PR #66, opened 2019-01-15
buildMobileVLCKit: write tools variables to config.mak

alexandre-janniaux:
Append tools variables to vlc/contrib/config.mak so that contrib can be build with the same configuration without setting the environment variables again.

This is the beginning of a set of changes  to ease the building of VLCKit. The main goal is to have the same building behaviour as other platforms.

The overall todolist is : 

* [ ] write configuration variable into the different config files used for building (either through ./configure for libvlc or the config.mak for the contribs)
* [ ] avoid using environment variable in the build script, as it might have an unexpected behaviour
* [ ] fix contribs that can build only because the configuration is in the environment variables
* [ ] fix passing options like tv_version_min when it shouldn't be (simulator builds)

### DIFF
--- a/compileAndBuildVLCKit.sh
+++ b/compileAndBuildVLCKit.sh
@@ -373,6 +373,13 @@ buildLibVLC() {
 
     echo "EXTRA_CFLAGS += ${EXTRA_CFLAGS}" >> config.mak
     echo "EXTRA_LDFLAGS += ${EXTRA_LDFLAGS}" >> config.mak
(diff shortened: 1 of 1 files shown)


fkuehne (approved):
First patch and intend of further patches sound good to me!

alexandre-janniaux:
It might take some time to do each step so I suggest that we fix each point in different MR.

carolanitz:
merged with b9d7ba67b090394452639b19df74103ab4b13a7b

PR #67, opened 2019-01-23
CircleCi: bump Xcode to avoid undefined symbols error

carolanitz:
undefined symbols for _SSLCopyALPNProtocols is caused because these headers are not implemented until 10.13.4 but SecureTransport.h incorrectly claims its 10.13.0

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ version: 2
 jobs:
   build:
     macos:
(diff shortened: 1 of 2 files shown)


fkuehne (approved):
Nice!

carolanitz:
merged with 3dd80ca65adfeca68bdf6cc1258e161417fbcc7d

PR #68, opened 2019-02-08
VLCRendererDiscoverer: fix crash in VLCRendererDiscoverer

carolanitz:
create the VLCRendererItem before we switch threads to avoid use after free


chouquette (approved):
LGTM :)

fkuehne (approved):
Looks good!

PR #69, opened 2019-02-18
VLCLibrary: VLCLibraryLogReceiverProtocol: Fix documentation

Mikanbu:
Fixes: https://code.videolan.org/videolan/VLCKit/issues/248

### DIFF
--- a/Headers/Public/VLCLibrary.h
+++ b/Headers/Public/VLCLibrary.h
@@ -130,8 +130,8 @@ NS_ASSUME_NONNULL_BEGIN
 @required
 /**
  * called when VLC wants to print a debug message
- * \param error the dialog title
- * \param message the error message
+ * \param message the debug message
+ * \param level the debug level
  */
 - (void)handleMessage:(NSString *)message
            debugLevel:(int)level;


Headers/Public/VLCLibrary.h
```diff
@@ -130,8 +130,8 @@ NS_ASSUME_NONNULL_BEGIN
 @required
 /**
  * called when VLC wants to print a debug message
- * \param error the dialog title
  * \param message the error message
+ * \param level the debug level
```
carolanitz, line 134: could you also change the error to debug for param message ?

carolanitz (approved):

Headers/Public/VLCLibrary.h
Mikanbu: I'm not sure to understand what you mean here?

carolanitz: ```param message the error message``` should be the debug message I think

Mikanbu: Ah indeed, I will change this!

PR #70, opened 2019-02-18
Buildscript: Fetch python3 path during buildLibVLC

Mikanbu:
Fetch python3 path during buildLibVLC since `TARGET` is set during buildLibVLC.


fkuehne (commented):
I like the idea of using a function here. However, the EXPORT should happen outside that function because otherwise it would be masked and hard to see that it's actually happening.

Mikanbu:
Indeed! I will submit a fix for this.

Mikanbu:
The issue has been addressed, if there are no other comments I will fixup everything and merge~

PR #72, opened 2019-03-14
transcode: fix transcode with passthrough video

alexandre-janniaux:
Fix transcode pipeline when using passthrough for the video track, for
example when using chromecast which doesn't try to convert the video
track.

Thus, it also fixes the playback of correct h264 stream on chromecast.

This was a regression from 553a5da7fe4f63066536ad5522bb0cf754ba8f1b
due to patch transcode-add-support-for-multiple-venc-parameters.


Mikanbu:
LTGM, it now works with samples provided by users. 🎉

carolanitz (approved):
after removing the commented code it looks good to merge :)

alexandre-janniaux:
If I understand correctly which comment you're talking about, I had left some commented code previously (no idea why, sorry for that) so it's `-+` in the patch of patch, but it's removed in this PR.

Mikanbu:
Merged, thank you for the PR. 🎉

PR #73, opened 2019-05-09
VLCMedia: Remove attempts to encode urls

carolanitz:
This removes bugs where urls were wrongly double encoded,
instead the App needs to ensure that the URL is percent encoded if needed

this fixes : https://code.videolan.org/videolan/VLCKit/issues/139
and https://code.videolan.org/videolan/VLCKit/issues/45


alexandre-janniaux:
LGTM, it simplifies the code and remove edge cases, and it seems better to me to always have real URL without the need to percent-encode them in VLCKit. The only client that can provide not-url-encoded strings is the final user itself, so making it a feature of the UI instead of the backend seems more rational to me.

Maybe adding documentation notes for the change can be another good candidate for this PR.

PR #74, opened 2019-06-21
VLCMediaListPlayer: Add libvlc callbacks

Mikanbu:
This adds libvlc event callbacks for media_list_player.

### DIFF
--- a/Headers/Public/VLCMediaListPlayer.h
+++ b/Headers/Public/VLCMediaListPlayer.h
@@ -3,11 +3,13 @@
  *****************************************************************************
  * Copyright (C) 2009 Pierre d'Herbemont
  * Partial Copyright (C) 2009-2013 Felix Paul Kühne
- * Copyright (C) 2009-2013 VLC authors and VideoLAN
+ * Copyright (C) 2009-2019 VLC authors and VideoLAN
+
  * $Id$
  *
  * Authors: Pierre d'Herbemont <pdherbemont # videolan.org>
- *          Felix Paul Kühne <fkuehne # videolan.org
+ *          Felix Paul Kühne <fkuehne # videolan.org>
+ *          Soomin Lee <bubu # mikan.io>
  *
  * This program is free software; you can redistribute it and/or modify it
  * under the terms of the GNU Lesser General Public License as published by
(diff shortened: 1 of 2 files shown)


carolanitz:
can you check why circle is complaining? Also from the description and commit message I have no background what exactly is being solved and why this is needed :/ is there a ticket or somewhere more background?

Sources/VLCMediaListPlayer.m
```diff
@@ -205,4 +273,28 @@ - (VLCRepeatMode)repeatMode
...
+#pragma mark - Delegate methods
+
+- (void)mediaListPlayerPlayed
+{
+    if ([_delegate respondsToSelector:@selector(mediaListPlayerPlayed)]) {
+        [_delegate mediaListPlayerPlayed:self];
+    }
+}
+
+- (void)mediaListPlayerNextItemSet:(VLCMedia *)media
```
carolanitz, line 287: if this is a set why is there only one media ?

Headers/Public/VLCMediaListPlayer.h
```diff
@@ -36,6 +36,13 @@ typedef NS_ENUM(NSInteger, VLCRepeatMode) {
     VLCRepeatAllItems
 };
 
+@protocol VLCMediaListPlayerDelegate <NSObject>
+@optional
+- (void)mediaListPlayerPlayed:(VLCMediaListPlayer *)player;
```
carolanitz: startedPlayback? or is this the callback for being done with playback. 
comments might help especially for other users of the sdk

Headers/Public/VLCMediaListPlayer.h
```diff
@@ -36,6 +36,13 @@ typedef NS_ENUM(NSInteger, VLCRepeatMode) {
...
+- (void)mediaListPlayer:(VLCMediaListPlayer *)player nextMedia:(VLCMedia *)media;
```
carolanitz: also here I'm unsure in what state we're in from the callback name

carolanitz (approved):
just add some comments update to 2019 add yourself as author and I think we're good here

Mikanbu:
Reviews addressed. 🎉

Sources/VLCMediaListPlayer.m
Mikanbu: The name is a bit confusing but this is mirroring the libvlc API.
We are exposing as `mediaListPlayer:nextMedia:` publicly.

carolanitz (approved):

Mikanbu:
Rebased and merged. Thanks for the reviews! 🎉

PR #78, opened 2020-03-09
Fixed rendererDiscovererItemAdded sending nil item

nihalahmed:
The callback was sending the rendererItem which is nil. It should be sending back the item which is not nil.


Mikanbu:
Thank you for this PR, this was merged by 06946f1e.

nihalahmed:
Thanks!

PR #81, opened 2021-01-25
3.0

alpertayfun:
Hello,

setUserAgent method in TVVLCKit ? 

Or can we add ? 

Thank you.


fkuehne:
There is `- (void)setHumanReadableName:(NSString *)readableName withHTTPUserAgent:(NSString *)userAgent;` on VLCLibrary since a long time.

PR #82, opened 2022-02-21
fixes https://code.videolan.org/videolan/VLCKit/-/issues/585

lechium:
libvlc_video_set_spu does not handle -1 as a call to turn off subtitles, this handles turning them off properly when -1 is passed as a parameter to match the documentation spec.


Mikanbu:
Since we prefer using our Gitlab instance, this PR has been superseeded by this MR: https://code.videolan.org/videolan/VLCKit/-/merge_requests/130

PR #84, opened 2022-05-28
Update issue templates

MarcelRaschke:


fkuehne:
Hello,

I like what you propose here. However, we don't use the GitHub issue tracker but prefer our own here: https://code.videolan.org/videolan/VLCKit/-/issues

However, your proposal could be easily adapted to fit gitlab's issue template system. A MR for that would be welcome!

Thanks!

PR #85, opened 2022-10-29
added sudo cause permission error

JonnycatMeow:


koushikcs562:
i'm interested in helping improve the build process and handling permission issues. My approach will be to carefully review the build scripts, identify where permission errors may occur, and suggest fixes like proper sudo usage or alternative solutions to make the build more reliable. I’ll open a PR shortly with my changes.

PR #87, opened 2025-09-24
Allow keyboard backlight to shut off during playback

pcl:
This change updates how the screen is kept on during video playback.

Previously, sleep was prevented by injecting artificial user interaction
on every timeChangeUpdate callback. This successfully keeps the display
from dimming and the computer from going to sleep, but it also prevents
the keyboard backlight from shutting off on laptops that have keyboard
backlights.

This commit uses IOPMAssertionCreateWithName to assert to the operating
system that the display should not sleep. It does this when playback
is detected, rather than on every timer update, and releases the assertion
when playback stops.

This results in new behavior in which the backlight will dim, based on
the user's systemwide backlight preferences.

It would be straightforward to preserve the previous behavior by default
and require a function call or configuration setting or something to 
enable the new behavior, if that is preferable. I believe that it's
generally a bug to keep the backlight on (and to pretend the user is
actually active) instead of using the NoDisplaySleep assertions. But
I don't have a sense for how the team generally handles these sorts
of behavioral changes.

Incidentally, in my app, I've hacked around this problem by monkey
patching the disableDelaySleep() function, which solves my problem,
but is clearly not the most elegant long-term solution.

### DIFF
--- a/Sources/Playback/VLCMediaPlayer.m
+++ b/Sources/Playback/VLCMediaPlayer.m
@@ -47,6 +47,7 @@
 #if !TARGET_OS_IPHONE
 /* prevent system sleep */
 # import <CoreServices/CoreServices.h>
+# import <IOKit/pwr_mgt/IOPMLib.h>
 /* FIXME: Ugly hack! */
 # ifdef __x86_64__
 #  import <CoreServices/../Frameworks/OSServices.framework/Headers/Power.h>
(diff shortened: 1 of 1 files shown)


pcl:
I've tested this out by manually building against the new dependency, and can see that it does the job in my app.

pcl:
Also, FYI, the API calls I'm using here date back to 10.7, whereas VLCKit is targeting 10.13 and higher. So there's no new version limitation here.

fkuehne:
Thanks a lot for this contribution!

I took it to our gitlab instance and will merge it shortly: https://code.videolan.org/videolan/VLCKit/-/merge_requests/378

PR #89, opened 2026-03-29
Closing this PR due to accidental creation

ci7lus:
sry


fkuehne:
Why not contribute your work?

ci7lus:
@fkuehne Thanks for reaching out.

These changes are currently very specific to my own project (handling Japanese ARIB MPEG-TS), and they aren't really ready for a general use yet. Some parts might even cause issues for other users.

I'm still refining the code, but once things settle down and I have something that’s stable and useful for upstream, I’ll look into contributing properly. Thanks again.

fkuehne:
As you are probably very well aware, there is basic ARIB MPEG-TS support in VLC, but not as detailed as done by you. This would be a very interesting addition indeed - same for what you added to FFmpeg.

PR #90, opened 2026-09-16
Retain media before dispatching player events

WhoSayIn:
With `VLCEventsLegacyConfiguration`, stopping or replacing media before the main queue drains can free the descriptor captured by a queued media callback. The metadata path can then crash in `objc_retain` while bridging `libvlc_media_get_user_data`.

Resolve or materialize a strong `VLCMedia` inside the synchronous libVLC callback, then capture only that wrapper in the dispatched block. Apply the same fix to media-changed, metadata, subitems, and artwork callbacks, which share the borrowed-descriptor capture. ARC also releases the capture when `VLCEventsHandler` drops an event because its weak target is gone. Event dispatch policy is unchanged.

Validation: 3,200 deterministic cases passed with Address Sanitizer using the extracted production callback bodies and the real Legacy configuration/event handler. The unpatched callback fails with a late-descriptor-lookup assertion. Coverage includes existing/new wrappers, null media, caller teardown before queue delivery, copied artwork, dropped events, main-thread asynchronous delivery, and zero surviving media/descriptor instances. The checked media doubles isolate this ownership contract; they do not substitute for real libVLC/device playback testing.

The metadata forwarding callback was introduced by ae3bbcea88b9f8a1076da3e4aef3fb829c234676.


WhoSayIn:
This was LLM generated and I let it create a PR because I am using vlckit myself and encountered a bug. I haven't seen anything in the Readme that was against AI generated PRs but let me know if that would be an issue and I'll make sure to not post them in the future :)

